PR #1, opened 2024-06-26
CI: adding HTML publishing workflow

bsipocz:
This is to debug the circleCI redirect workflow for PRs as the config below should not run on PRs


bsipocz:
I don't really see why this run into a DeadKernelError, locally the htmls build without an issue (there are 12 warnings about headings, but we see those in the circleCI log, too before the kernel error)

PR #2, opened 2024-06-26
CI: not yet fail builds with warnings. And minor config updates

bsipocz:


bsipocz:
CircleCI fails, but the same job runs until the end in the publish job. Not sure what is going on, but postpone any further investigation for now.

PR #10, opened 2024-07-31
CONT: adding firefly tutorials

bsipocz:


bsipocz:
Both these notebooks work for me locally on OSX, so I'm puzzled if the issues we see in CI are expected or not.

cc @jaladh-singhal in case you have any tips or insights of what is going wrong.

jaladh-singhal:
> cc @jaladh-singhal in case you have any tips or insights of what is going wrong.

@bsipocz this `gaierror: [Errno 8] nodename nor servname provided, or not known` has been observed by me, Eric, and others randomly - it's hard to reproduce consistently. It's interesting to see that it's happening only on MacOS pipeline. 

A quick google search and chasing SO threads, reveals that it has to do with requirement of explicit defining of hosts/port name for MacOS (but I can be wrong). I'll open a bug ticket about it at firefly-client so that we can remember to debug it.

tutorials/requirements.txt
```diff
@@ -10,5 +10,9 @@ pandas>=1.5.2
 dask[distributed]
 psutil
 s3fs
+firefly_client
+jupyter_firefly_extensions
 # For supporting myst-based notebooks
 jupytext
+# For firefly-client packaging bug
+setuptools
```
jaladh-singhal, line 18: Is this something we need to fix at package configuration of firefly-client by defining setuptools as required dependency?

tutorials/requirements.txt
```diff
@@ -10,5 +10,9 @@ pandas>=1.5.2
 dask[distributed]
 psutil
 s3fs
+firefly_client
+jupyter_firefly_extensions
```
jaladh-singhal, line 14: Couldn't help but notice: the pip packages names for both of these use `-` instead of `_` - I ran into similar confusion earlier and couldn't install the package, I'm surprised to see it's working in CI with `_`

tutorials/requirements.txt
bsipocz: yeap, see https://github.com/Caltech-IPAC/firefly_client/issues/60, as well as the other `pkg_resources` related issue. In fact addressing Tim's would make this one irrelevant, too.

tutorials/requirements.txt
bsipocz: hahh, interesting. I suppose pip replaces them at install, but indeed I should declare them as they are named on pypi.

bsipocz:
> @bsipocz this gaierror: [Errno 8] nodename nor servname provided, or not known has been observed by me, Eric, and others randomly - it's hard to reproduce consistently. It's interesting to see that it's happening only on MacOS pipeline.

It seems to be pretty consistent on CI, but I also don't see it locally on OSX, everything works there for me when I directly use the notebooks. 
So, is there a workaround recipe that you would suggest us to do here?  I mean, the easiest would be to just exclude these tutorials from the OSX testing for now and comment in a reference for a known firefly issue.

tutorials/requirements.txt
bsipocz, line 14:
```suggestion
firefly-client
jupyter-firefly-extensions
```

tutorials/requirements.txt
jaladh-singhal: Ah, didn't notice that - thanks. I'll get this added to Jira

jaladh-singhal:
> So, is there a workaround recipe that you would suggest us to do here? I mean, the easiest would be to just exclude these tutorials from the OSX testing for now and comment in a reference for a known firefly issue.

Yes, I would suggest excluding these (or anything under `firefly/`) for now. I have other stuff on my plate so I won't get chance to work on it until late August

bsipocz:
> Ubuntu and Windows pipelines passing is good enough check, I think?

Well, not if we want to ensure things are working for mac users, too, but for now I'll indeed turn off that job for that directory.

PR #27, opened 2024-08-29
HTML: restructure index page headings

bsipocz:
This also adds top level headings as captions to the sidebar. Downside of this workaround is that those are not linked to the achors, so it's not possible to use them to navigate back. It also won't work once we add one more layer of headings in the index page (like it was in the first draft, before this PR)

I saw some other workarounds, so can keep experimenting if we


bsipocz:
Hmm, `caption` doesn't work as it shows up in the main index page and not just on the sidebar, too. So back to square one.

bsipocz:
Brute force `sed` saved the day, a 4th workaround idea to edit the rendered result after the fact actually works. I think it's best to keep this around until upstream comes up with an actual solution.

PR #28, opened 2024-08-30
CI: fix 'html rendering' test skip logic

bsipocz:
Fix bug in the skip logic


bsipocz:
Merging as it triggered the jobs, the actual run results don't matter here.

PR #16, opened 2024-08-19
Add notebooks for NEOWISE parquet catalog

troyraen:
Adds two notebooks that demo access to the parquet version of NEOWISE Single-exposure Source Table:

- neowise-source-table-strategies.md
- neowise-source-table-lightcurves.md

PR #33, opened 2024-09-25
CI: add optin to build html on GHA

bsipocz:
The need for this came up in https://github.com/Caltech-IPAC/irsa-tutorials/pull/16 where the newly added content runs out of circleCI resources, but it's not clear whether it would do the same on GHA (as other notebooks execute without issues on GHA but not on circleCI)


bsipocz:
OK, this works as expected, so I go ahead and merge it now.

Issue #15, opened 2024-08-09
`multiprocessing` considerations


troyraen:
Testing the NEOWISE parquet catalog notebooks ([IRSA-6202](https://jira.ipac.caltech.edu/browse/IRSA-6202), https://github.com/IPAC-SW/ipac-sp-notebooks/pull/69) revealed an issue with python's `multiprocessing` (thanks to Chris Gelino for running into it) that affects other notebooks in this repo and also the rendered pages. It's a bit involved so I'll start with the summary.

## Summary

Notebooks in this repo that use python `multiprocessing` for parallelization will not work out of the box on Mac or Windows. They'll work just fine on Linux, which includes both Fornax and IRSA science platform. Conversely, our notebooks in fornax-demo-notebooks that use `multiprocessing` should work out of the box everywhere because we made a different design choice in that repo. The tradeoff is that Fornax notebooks will be more difficult to serve.

@bsipocz and I talked this over and agree those are probably the right choices for both repos, at least for now. We'll see how it goes both ways and can adjust if needed. Posting here so others are aware of it for other notebooks and also to have a larger discussion if needed.

## The `multiprocessing` problem

Linux and Mac have different defaults in how child processes are started and unfortunately the Linux default, which works great for these notebooks, is not exactly reproduced on a Mac even when choosing the same option. Specifically, the problem on Mac is that the functions defined in the notebook (which runs in the main process) do not get copied to the pool workers (child processes). So the multiprocessing cell throws an `AttributeError` complaining that it can't find the function it's supposed to run. The easiest solution is to save the functions in a separate .py file and import them. Then the workers can find them. 

The docs say that the Linux default option is not even available on Windows. I don't have a Windows machine to try with, but my best guess is that the same solution will be best there as well (though I never quite know with Windows).

## The serving problem

Unfortunately, the solution of defining functions in a separate .py file collides with our preference that notebooks in this repo avoid relying on ancillary files whenever possible. This is because ancillary files are difficult to serve. -- The process of notebook rendering will work fine, but the public webpages will not show the ancillary files. We'd have to leave instructions for users to manually download them from the repo. @bsipocz can say more here.

We made the opposite choice for the fornax-demo-notebooks repo where all of the notebooks have ancillary .py files that store most of the code. Those notebooks require a whole lot more code than notebooks in this repo, at least so far. So I think the separate files there makes a lot of sense because those notebooks would be so huge if they included all the code directly.

troyraen:
[...]

PR #34, opened 2024-09-27
Patch Parallelize_Convolution.md

troyraen:
- Add a cell to install dependencies if needed.
- Add `multiprocessing` note for Mac and Windows users. Addresses #15 for this noteboook.
- Bugfix dask call. This is a repeat of IPAC-SW/ipac-sp-notebooks/pull/42 which apparently didn't get ported over.


bsipocz:
Thank you! This should not fail CI, so I'll come back to it for a deeper dive to see what actually goes wrong.

bsipocz:
> This should not fail CI,

Never mind, all the failures are due to the timeouting cosmodc2 TAP query. That one feels to be especially flaky.

bsipocz (approved):
Thanks! All looks good, and CI failure is a red herring here.

PR #35, opened 2024-09-27
DOC: making pip install cells consistent

bsipocz:
@troyraen has highlighted in #34 that we haven't yet done this consistency cleanup here. This is purely a cleanup of uncommented lines, so no need to run the tests.


troyraen (approved):
Thanks!

Issue #20, opened 2024-08-20
BUG: links should be links in Firefly SED notebook


bsipocz:
Currently, the links are normal text rather than working URLs (e.g. link to the paper being referenced, or a firefly link)

bsipocz:
This has been addressed

PR #40, opened 2024-12-17
Add OpenUniverse2024 Firefly notebook and update other firefly notebooks

jaladh-singhal:
- Adds OpenUniverse notebook
- Updates in other 2 firefly notebooks
  - Homogenize how to instantiate firefly-client
  - API update in SEDs notebook
  - More descriptive headings in NEOWISE notebook
- Adds FIREFLY_URL to env (just in case) as fallback

### DIFF
--- a/tutorials/firefly/OpenUniverse2024Preview_Firefly.md
+++ b/tutorials/firefly/OpenUniverse2024Preview_Firefly.md
@@ -67,6 +67,8 @@ https://github.com/Caltech-IPAC/jupyter_firefly_extensions/blob/master/README.md
 - firefly_client.FireflyClient for using the Firefly python client
 - astropy.nddata.Cutout2D for making image cutouts
 - itertools.product to support looping over Roman blocks
+- reproject.reproject_interp to convert Roman coadds from STG to TAN projection
+- io.BytesIO for writing a fits file to an in-memory stream
 
 ```{code-cell} ipython3
 from astropy.io import fits
@@ -80,6 +82,8 @@ from astropy.coordinates import SkyCoord
 from firefly_client import FireflyClient
 from astropy.nddata import Cutout2D
 from itertools import product
+from reproject import reproject_interp
+from io import BytesIO
 ```
 
 ## Learn where the OpenUniverse2024 data are hosted in the cloud.
@@ -565,7 +569,7 @@ For each row in the table you can notice a marker in the image. Selecting a row 
 
 ## Show only high-redshift galaxies
 
-High redshift galaxies are the most interesting, so let's filter the table we sent to Firefly to only include z>3 galaxies. Notice how the table display and image overlay change. WHAT HAPPENS TO THE CHARTS? You can remove this filter or add new ones through the GUI.
+High redshift galaxies are the most interesting, so let's filter the table we sent to Firefly to only include z>3 galaxies. Notice how the table display and image overlay change. Notice how the chart becomes a scatterplot from a heatmap because the sources reduce. You can remove this filter or add new ones through the GUI.
 
 For filtering, we will use `apply_table_filters` method on the galaxy table we loaded above. More info about this method: https://caltech-ipac.github.io/firefly_client/api/firefly_client.FireflyClient.html#firefly_client.FireflyClient.apply_table_filters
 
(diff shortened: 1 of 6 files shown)


bsipocz:
Ahh, I'm not sure how we should proceed with testing this notebook. We didn't have any issues with the previous firefly ones, but this one throws an error complaining about being outside of jupyter:

```
RuntimeError: FireflyClient.makeLabClient can only be used in the Jupyterlab environment. fix jupyter_firefly_extensions in Jupyter Lab or use FireflyClient.make_client()
```

bsipocz:
I would say, let's add it to the rendering list, maybe jupyter-book will be enough for it to be picked up and it's only picky about pytest

jaladh-singhal:
> Ahh, I'm not sure how we should proceed with testing this notebook. We didn't have any issues with the previous firefly ones, but this one throws an error complaining about being outside of jupyter:
> 
> ```
> RuntimeError: FireflyClient.makeLabClient can only be used in the Jupyterlab environment. fix jupyter_firefly_extensions in Jupyter Lab or use FireflyClient.make_client()
> ```


@bsipocz the other notebooks use `make_client()` for this very reason. I can revert it to that, but I was wondering if it's using latest notebook 7 (as opposed to notebook classic), the extension should work. I can't track the version number of notebook, is it not 7?

jaladh-singhal:
Oh and installing the extension in enviroment should already install jupyter lab: https://github.com/Caltech-IPAC/jupyter_firefly_extensions/blob/master/pyproject.toml#L2-L5

So, is the testing pipeline not running the notebooks in notebook/lab context, but as a python script?

bsipocz:
> So, is the testing pipeline not running the notebooks in notebook/lab context, but as a python script?

Yes, the tests are running them as python code using pytest. We use something different for the HTML builds (and execution), so my hope is that will just work out of the box here.

bsipocz:
Unfortunately, I see the error with the rendering, too, so it doesn't like either of our machinery.

bsipocz:
I'm working on the binder integration parallel with this PR, and needed to reshuffle the config files: please rebase so you can add the changes to the requirements file at the correct location.

tutorials/firefly/OpenUniverse2024Preview_Firefly.md
```diff
@@ -0,0 +1,750 @@
...
+
+Firefly is an open-source web-based UI library for astronomical data archive access and visualization developed at Caltech and used by multiple space- and ground-based astrophysics archives. More information can be found here:
+https://github.com/Caltech-IPAC/firefly/blob/dev/README.md
+
+In addition to being used to make web applications, Firefly can be used from Python. More information can be found here:
+https://caltech-ipac.github.io/firefly_client/usage/index.html
+
+The Firefly JupyterLab Extension makes it particularly easy to use Firefly to efficiently visualize cloud-hosted astronomical data using cloud-based JupyterLab instances. More information can be found here:
+
+https://github.com/Caltech-IPAC/jupyter_firefly_extensions/blob/master/README.md
```
bsipocz, line 53: These don't render as URLs, so they need to be fixed.

tutorials/firefly/OpenUniverse2024Preview_Firefly.md
```diff
@@ -0,0 +1,750 @@
...
+
++++
+
+## Imports
```
bsipocz, line 57: I don't think we need to go into this much details for each submodules, but since these are here I don't advocate for their compression, only if the scientists would ask for it, too.

tutorials/firefly/OpenUniverse2024Preview_Firefly.md
```diff
@@ -0,0 +1,750 @@
...
+- astropy.wcs for dealing with astronomical world coordinate systems
+- astropy.units for dealing with astronomical units
+- astropy.coordinates.SkyCoord for dealing with astronomical coordinates
+- firefly_client.FireflyClient for using the Firefly python client
+- astropy.nddata.Cutout2D for making image cutouts
+- itertools.product to support looping over Roman blocks
+- reproject.reproject_interp to convert Roman coadds from STG to TAN projection
+- io.BytesIO for writing a fits file to an in-memory stream
+
+```{code-cell} ipython3
```
bsipocz, line 73: A cell with a commented out `pip install ` content is missing. Please add that in the same style as the other notebooks have (commented out list of all the directly imported dependencies. If firefly require something else that is not directly imported but is not installed as a dependency of something, then list those, too)

tutorials/firefly/OpenUniverse2024Preview_Firefly.md
```diff
@@ -0,0 +1,750 @@
...
+             Title="Rubin Coadd"
+             )
+```
+
+### Use ds9 region syntax to overplot the simulated Roman image blocks on the interactive display
+
+The Firefly client includes several methods related to controlling ds9 region overlays. To 
+overlay a region layer on the loaded FITS images, we can use `overlay_region_layer`:
+
+https://caltech-ipac.github.io/firefly_client/api/firefly_client.FireflyClient.html#firefly_client.FireflyClient.overlay_region_layer
```
bsipocz, line 409: these look really ugly, please make them a link instead with a shorter text. 

(I'm not certain that interpshinx linking would work out of the box, we can open an issue for it and I can look into it in the next year; for now manually linking these is  way to go)

bsipocz (approved):
Overall looks good. I haven't looked into the code itself only the rendering. 

My comments are mostly about making links and making the look nicer. That is certainly suitable for a follow-up, and we already have a somewhat relevant #20 for it.


The more critical comment, that is certainly requires a follow-up PR, is to add a pip install cell that lists all the dependencies. Think about it as a way of making the notebook a standalone, someone with a minimal python install and jupyter should be able to run the notebook once they install the dependencies by uncommenting that cell.

bsipocz:
Thank you @jaladh-singhal!

tutorials/firefly/OpenUniverse2024Preview_Firefly.md
jaladh-singhal: Hmm, I didn't check the html build until now, they were rendering in the notebook.

tutorials/firefly/OpenUniverse2024Preview_Firefly.md
jaladh-singhal: Yes, I'll defer it to Vandana

tutorials/firefly/OpenUniverse2024Preview_Firefly.md
jaladh-singhal: Oh, ok - I'll add it

tutorials/firefly/OpenUniverse2024Preview_Firefly.md
jaladh-singhal: Agreed, they all are looking ugly. I thought @vandesai1 kept it in this formatting so that audience can know and click it. I can change it to hyperlinked method name?

jaladh-singhal:
@bsipocz Thank you for review and merging it on such a short notice. I'll open up a follow up PR for these text fixes. And once you're back let me know if you have any request for code changes, I'll add them to that PR too (hopefully in the week before AAS)

tutorials/firefly/OpenUniverse2024Preview_Firefly.md
bsipocz: yes, I think hyperlinked method, or class+method name is the way to go, just like as it would look like in a sphinx API docs

PR #42, opened 2025-01-07
Add a Contributor Guide

### LINKED ISSUE #17: DOC: contributing and usage guides
We will need to write up some short documents for 1) contributing to this repo; as well as 2) bulk usage; e.g. cloning the whole repo either into a SP environment, or directly by the users (as opposed to pick and choose from the HTML renderings).

In order to keep the repo clean of random branches, I would suggest following a fork based workflow for contributing.

jaladh-singhal:
Aims to fix #17 esp the bulk usage part

I've tried to  draft a documentation based on my development experience on this repository so far. It's very minimal and can use a lot of changes (and maybe restructuring too?)

Marking it as draft since it needs feedback from other developers (@bsipocz, @troyraen, ...?) before adding more content

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,34 @@
+# Contributor Guide
+
+## Environment setup and workflow
+- Clone the repo
+- Create a fresh python environemt with `.binder/requirements.txt` file
+- Open `jupyter lab` or `jupyter notebook` (so that jupytext extension can take effect)
+- Then right-click on a MyST markdown notebook, choose `Open With` -> `Notebook`
+or `Jupytext Notebook` (more info [here](https://nasa-navo.github.io/navo-workshop/00_SETUP.html#handling-notebooks-in-myst-markdown-format)).
+
+> Note: If you'd like to work with this repository in an editor/IDE of your choice,
+make sure that you have relevant extensions to parse MyST markdown notebooks to
+interactive python notebook (ipynb) and to reflect the changes made there to markdown
+format:
+>- VSCode: doesn't seem to have a reliable extension to open md files as notebook, so you may have to manually pair them using jupytext in terminal.
+>- Other editors: ...
+
+## Updating an existing notebook (.md format)
+- Open it in jupyter notebook or lab, jupytext extension should be installed (it's already listed in `requirements.txt` file mentioned above)
+- Right click on a md notebook file and "open with" notebook format. 
+- Edit it as you would a regular notebook and once saved the changes should 
+reflect in md
+- Make sure only content changes appear in md diff, not the frontmatter, before 
+comitting
+- Make a PR once ready (TODO: explain how like forking etc.)
+
+## Adding a new notebook (from .ipynb format)
+To add an ipynb notebook you were working on to this repository, convert that to
+MyST markdown using:
+
+```python
+jupytext --to md:myst notebook.ipynb 
+```
+
+Then add it to `tutorials/` and also add its path in a relevant TOC in `index.md`.


CONTRIBUTING.md
```diff
@@ -0,0 +1,34 @@
+# Contributor Guide
+
+## Environment setup and workflow
+- Clone the repo
+- Create a fresh python environemt with `.binder/requirements.txt` file
```
jaladh-singhal, line 5: @bsipocz I noticed you recently moved requirements.txt to `.binder/` from `tutorials/`. Can we keep it in `tutorials/` and create a symlink to it in `.binder/`?

bsipocz: Symlinks notoriously won't work on windows, at least I had issues with keeping symlinks in repos before (and we only symlinked licences and contributing guides rather than actually important content)

CONTRIBUTING.md
```diff
@@ -0,0 +1,34 @@
...
+- Open `jupyter lab` or `jupyter notebook` (so that jupytext extension can take effect)
+- Then right-click on a MyST markdown notebook, choose `Open With` -> `Notebook`
+or `Jupytext Notebook` (more info [here](https://nasa-navo.github.io/navo-workshop/00_SETUP.html#handling-notebooks-in-myst-markdown-format)).
```
bsipocz, line 8: we should link more upstream than NAVO, e.g. numpy

bsipocz (commented):
I'm not yet sure if this is the right repo for this document, or we should rather keep it in the other repo (after all that's where we have the template, too).

Anyway, I'll give this a more detailed review later this week.

CONTRIBUTING.md
jaladh-singhal: Can you share the link? Last I checked numpy docs, I got confused because they were instructing to "pairing the notebook and md" but then I learned from you that that's not required.

bsipocz: I knew it would bite me back, we need to cleanup the numpy docs first, 😅 : https://github.com/numpy/numpy-tutorials/issues/186

CONTRIBUTING.md
troyraen: I thought users wouldn't need to touch this requirements.txt file all since notebooks in this repo should always `pip install` what they need in the first code cell -- I thought this file was only here for the rendering? I do see that the instructions below have some dependency on in it (i.e., assuming jupytext extension is installed). Since my workflow with notebooks is so different I don't know exactly what the user will need, so I'm fine with whatever's easiest (either telling them to install that extension separately or just pointing them to the requirements.txt file even though it isn't intended for them). Although, the other repo doesn't have a requirements.txt, so if this file gets moved there that's another factor to think about.

CONTRIBUTING.md
```diff
@@ -0,0 +1,34 @@
...
+- Make sure only content changes appear in md diff, not the frontmatter, before 
+comitting
+- Make a PR once ready (TODO: explain how like forking etc.)
+
+## Adding a new notebook (from .ipynb format)
+To add an ipynb notebook you were working on to this repository, convert that to
+MyST markdown using:
+
+```python
+jupytext --to md:myst notebook.ipynb
```
troyraen, line 31:
```suggestion
```sh
jupytext --to md:myst notebook.ipynb 
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Adding a new notebook (from .ipynb format)
+To add an ipynb notebook you were working on to this repository, convert that to
+MyST markdown using:
+
+```python
+jupytext --to md:myst notebook.ipynb 
+```
+
+Then add it to `tutorials/` and also add its path in a relevant TOC in `index.md`.
```
troyraen, line 34: Just flagging this line because there are similar instructions in the existing docs in the other repo, but the process is different in that repo. So we need to sync them up depending on which repo we chose.

troyraen (commented):
Thanks @jaladh-singhal, this is helpful. I don't often work in a jupyter environment (I'm much more likely to convert the .md to .py and just use an editor), so I don't really have thoughts/opinions about the actual instructions here except to say that I think they're much needed since contributors are likely to be working this way.

Organizationally, we should merge this with the template @bsipocz referred to so that we have one consistent set of "contributing" documentation. That template is for new notebooks and is here: https://github.com/IPAC-SW/ipac-sp-notebooks/tree/main/template. 

In terms of which repo, we had talked about having contributions go into the other repo both so that they can be worked on in private and so we can better control the content in this repo (since the rendering, etc. means there are more requirements/constraints here). But I'll defer to @bsipocz on that since she does most of the management tasks that would be affected.

In terms of the content, there's a little bit of overlap so we should make sure those parts are consistent (and this will depend on which repo we choose). Also, maybe the content in the template's README should actually go in the 'new notebook' section of this file. Then we wouldn't need that separate file. @jaladh-singhal take a look at that readme and the template and see what you think.

PR #44, opened 2025-02-19
Adding Euclid skeleton

bsipocz:
Just the directory for now, when we have deployable content, they should be added to the rendered pages, too.


bsipocz:
Merging this without content as the new notebooks are going into the sandbox first, in prep for the release.

PR #45, opened 2025-02-21
ENH: switching to use async TAP cosmodc2

### LINKED ISSUE #43: MAINT: try switching the cosmodc2 notebook to use async TAP
The cosmodc2 TAP notebook is frequently runs into timeout errors when run in CI (see e.g. https://github.com/Caltech-IPAC/irsa-tutorials/actions/runs/13363252210 where it timeouted 3 out of the 5 jobs. But sometimes it passes all 5).
The example is using `service.search` which defaults to `run_sync`. 

I wonder if this is a good point to showcase async TAP queries, and if that would resolve the timeouts? (The latter cannot really be directly tested as it's not exactly deterministic when to fail)

cc @vandesai1 as the author of the notebook.

bsipocz:
My expectation for this to fix the frequent timeouts.

TODO: change the narrative docs to add a sentence or two about sync and async. I'll do that once this cleans CI.

Close #43

### DIFF
--- a/tutorials/cosmodc2/cosmoDC2_TAP_access.md
+++ b/tutorials/cosmodc2/cosmoDC2_TAP_access.md
@@ -75,7 +75,7 @@ adql
 
 ```{code-cell} ipython3
 # In order to use TAP with this ADQL string using pyvo, you can do the following:
-result = service.search(adql)
+result = service.run_async(adql)
 result
 ```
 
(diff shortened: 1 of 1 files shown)


bsipocz:
OK, the notebook in question has passed the tests while it is now consistently failing on main (the run I triggered at the same time: https://github.com/Caltech-IPAC/irsa-tutorials/actions/runs/13452266524). So I'll go ahead with the text updates and will get this merged.

bsipocz:
Tests pass, so I go ahead with the merge now. If any more wording is needed around async/sync we can always add it in a follow-up.

PR #49, opened 2025-03-13
ENH: Euclid notebooks

bsipocz:
This PR adds the Euclid notebooks. They are still rely on VPN accessible irsadev, but we'll flip the URL prior to the release.

I'll try to do cleanups in incremental logical commits, but am also happy to go ahead and merge  this and do follow-up PRs each by each notebooks.

Some outstanding issues I see in the first round:

- [ ]  no reason to use pyvo, we can switch to astroquery
- [ ] no reason to do flux conversions by hand, we should use astropy
- [ ] we really could use more of astropy Table, none of pandas functionality I saw is specific so it's a bit counterintuitive to go back and forth.
- [ ]  direct use of requests and bytesIO feels a bit overkill, I think we will be able to remove it
- [ ]  there are some hard wired path pointing to personal home directory. That should be removed, even if it's just listed in a commented out cell.
- [ ]  we could/should remove the Exercise sections as those are not exactly appropriate for these types of tutorials (as opposed to hand-on workshops)
- [ ]  Code should be pep8-ed
- [ ]  add the whole thing into rendering
- [ ]  add the whole thing into CI
- [ ] revise headings

Some headings feel duplicate with the notebook title, and I would suggest their removals.


bsipocz:
OK, so the headings will definitely need some adjustments as everything is doubled in the rendering TOC:

Screenshot 2025-03-12 at 23 39 15

bsipocz:
I would say we go ahead and merge this now as is, and address the checklist as separate follow-up PRs. E.g. most importantly I want to have a separate one for rendering and other prettify-ing questions. While that is all separate from the code improvements.

bsipocz:
One new thing I noticed is that the headings are not actual headings but more like narrative sentences. And that won't really work, we need to threat the headings as titles for the given sections or paragraphs as they will be harvested for TOC.

A good example is here: https://caltech-ipac.github.io/irsa-tutorials/tutorials/cloud_access/cloud-access-intro.html, see the TOC on the right hand side, that is from the headings, so we'll need something concise and summarizing.
Some of the `####` level ones feels to be more like a note or warning, if needed we can put a box around them.

Anyway, I fix the header level inconsistency warnings in here, and we can take the rest in follow-ups.

bsipocz:
An example is here: I added a note instead of a level 4 heading, but you see how busy the TOCs are

bsipocz:
OK, no point in waitig for CI here as these are all excluded from the testing for now and the label seems to be broken.

PR #50, opened 2025-03-13
Add a Euclid cloud access notebook

jaladh-singhal:
Added a notebook to demonstrate cloud access to Euclid data (taking inspiration from Tiffany's recent notebooks)

PR #52, opened 2025-03-14
Euclid rendering

bsipocz:
This PRs primary purpose is to add the new notebooks to the HTML pages. It will be merged just before the data release, so ideally should be independent of any improvements done to the notebooks, so we can merge those sooner.

I'll also rebase it on top of #50, so the cloud notebook is included in the rendering, too.

(Note: execution of the code is turned off here. If we want to deploy any outputs those have to be generated locally on a computer with VPN access and then pushed directly to the deploy branch)

### DIFF
--- a/.binder/requirements.txt
+++ b/.binder/requirements.txt
@@ -17,6 +17,7 @@ firefly-client>=3.2.0
 jupyter-firefly-extensions
 reproject
 photutils
+fsspec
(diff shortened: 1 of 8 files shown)


bsipocz:
Failure is unrelated as it's a cosmoDC2 timeout. Merging this now.

PR #59, opened 2025-03-19
Adding Euclid notebook4

bsipocz:
This notebook could use some more narrative text, so we are not shipping it out just yet


bsipocz:
@tmesh - Adding you updates from the other repo. 

Once this passes all tests we may consider merging and making the notebook public, even though I feel the narrative texts are still missing or very thin in this one.

bsipocz:
Well, this notebooks also runs out of RAM, so need some benchmarking/revisions.

(RAM is 8GB, while all the network traffic is 3.3, so we likely keep too many copies around)

bsipocz:
(ignore CI statuses, everything was green in the previous run and we have ended up hitting IRSA blocktime with this latest one. All relevant check for this euclid notebook passed, status is coming from other notebooks)

PR #61, opened 2025-03-19
CI: Adding notebooks to the CI

bsipocz:
Most of the notebooks can be added to the CI. For now I leave out the two that runs into troubles with resources at rendering.


bsipocz:
docs build failure is unrelated cosmoDC2 timeout.

PR #67, opened 2025-03-20
CI: enable euclid notebook execution for HTML building

bsipocz:


bsipocz:
Merging as this was passing before the commit that removed the warnings. Note: notebooks 1 and 5 are running out of CI resources and thus cannot be plugged in here as is.

PR #71, opened 2025-03-24
A bit extra markdown in the second notebook

tmesh:


tutorials/euclid_access/2_Euclid_intro_MER_catalog.md
```diff
@@ -92,14 +92,22 @@ for col in columns:
     print(f'{f"{col.name}":30s}  {col.unit}  {col.description}')
 ```
 
+## Note that the mer catalog contains 476 columns, below are a few highlights:
```
bsipocz, line 95: We need to cleanup these notes, currently it's added as a level2 heading which doesn't seem correct; I would advise using the `Note` or `Tip` directives/admonitions instead:

https://mystmd.org/guide/admonitions

PR #74, opened 2025-03-25
CI: test only diffed notebooks

bsipocz:
We should not hit the servers unnecessarily

(included some code changes here, too, so we eventually have some test diffs in this PR, but all those need to be reverted before merging)


bsipocz:
Wow, this works nicely. Now the only thing to solve is the HTML caching, so we would only execute the diffed notebooks and use the rest from the cache.

PR #81, opened 2025-03-26
Updated so MER images plotted as cutouts

tmesh:
updating notebook 4 to be a cutout of the MER image, updated the adql query, updated the author list

### DIFF
--- a/tutorials/euclid_access/4_Euclid_intro_PHZ_catalog.md
+++ b/tutorials/euclid_access/4_Euclid_intro_PHZ_catalog.md
@@ -45,7 +45,7 @@ The photometry of every source is processed through a photometric redshift fitti
 
 ```{code-cell} ipython3
 # Uncomment the next line to install dependencies if needed.
-# !pip install requests matplotlib pandas astropy pyvo fsspec firefly_client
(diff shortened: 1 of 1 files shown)


tutorials/euclid_access/4_Euclid_intro_PHZ_catalog.md
```diff
@@ -260,14 +257,39 @@ df_g_irsa['x_pix']=xy_irsa[0]
 df_g_irsa['y_pix']=xy_irsa[1]
 ```
 
+Due to the large field of view of the MER mosaic, let's cut out a smaller section (30'x30')of the MER mosaic to inspect the image
```
bsipocz, line 260: why not use the astropy cutout?

bsipocz (approved):
I think this can still use a lot of cleanup, but the notebook works and renders without an error, so I go ahead and merge it.

Some comments: the object we pull the spectra for has changed, and the new one looks kind of flat. Is that change intended? (the previous one, besides the obvious noisy spikes, looked to have some feature around 1.35 microns.

mpl understands astropy and its units, so I'll add that to the list of cleanups

bsipocz:
Thanks!

PR #85, opened 2025-03-28
Euclid: Simplify spectra notebook

bsipocz:
This PR:

 - swaps PyVO to astroquery
 - uses astropy API consistently
 - cleans up pandas usage as there is nothing specific it is required for
 - relies on astropy Quantity support for plotting
 - removes usage of BytesIO as `fits.open` is fully capable of dealing with URLs
 - Cleans up the narrative a little bit to make code content and text more consistent

Please check the rendered HTMLs for an easier view.

PR #89, opened 2025-03-31
Meshkat Notebook 3 updates

tmesh:
Updated notebook 3 by adding a few imports, changing the target (different object id), removing "AND uri IS NOT NULL" as Anastasia says that isn't necessary when we have the object_id, updating the primary plot (with colors, error, legend), adding a plot of the MASK column and adding a description of some of the columns.


bsipocz:
Changes from here are included in #85; as well as the cleanup of these changes, so plotting is done in 3 lines without unnecessary loops and value conversions.

PR #90, opened 2025-04-01
BUG: avoid QTable convertint int columns to float

bsipocz:
We don't plot from `result_table`, so we don't strictly require it to be a `QTable`. 

So, this is a reasonable workaround while the upstream issues https://github.com/astropy/astropy/issues/17963 and the `NA` unit from the DB side are resolved.

Also, note that we see the values changed in the table during the conversions, which I still need to investigate and open relevant upstream bug reports for.


bsipocz:
Merging this as tests passed and I expect rendering to work for the changes.

PR #92, opened 2025-04-01
CI: adding caching for notebooks

bsipocz:
Opening it as a draft as I doubt this will work at first try (and I don't really have any working examples in front of me for this).


bsipocz:
This does the trick for caching on circleCI, through there are still some potential issues with being able to update the cache.

PR #93, opened 2025-04-08
ENH: adding roman simulation density tutorial

bsipocz:
I add this notebook to the pages now, but I would like to have the thumbs up from @jkrick and @vandesai1 about e.g the placing, etc.

Outstanding todos I noticed so far, but we should add to the list as they come up: 

 - [ ] rename download directory to `data` to have a bit more agreement amongst tutorials (this maybe a more overarching issue as the others may not be consistent yet either)
 - [ ] We may want to say a bit more about the dataset, and link to the IRSA website of it, etc.

### DIFF
--- a/.binder/requirements.txt
+++ b/.binder/requirements.txt
@@ -1,4 +1,6 @@
-# For the content of the tutorials
+# For the content of the tutorials, these packages are used directly or are
+# undeclared optional dependencies of functionalities we rely on
+#
 tqdm
 numpy>=1.24
 matplotlib>=3.7
@@ -19,6 +21,8 @@ reproject
 photutils>=2.0
 fsspec
 sep>=1.4
+h5py
+requests
 # For supporting myst-based notebooks
 jupytext
 # Making admonotions look nice for the myst notebooks
--- a/index.md
+++ b/index.md
@@ -91,6 +91,15 @@ tutorials/firefly/OpenUniverse2024Preview_Firefly
 
 ```
 
+```{toctree}
+---
+maxdepth: 1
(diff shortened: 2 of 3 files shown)


jkrick (approved):
I was thinking this notebook would fit under "Accessing IRSA’s on-premises holdings using VO protocols"

bsipocz:
@jkrick - but it doesn't use VO protocols but the data dump in h5py files.

bsipocz:
@jkrick  - Hmm, I've just looked at the rendered plots and the last jackknife one is flat while the caption claims otherwise. Is that expected? I don't see any errors hit by the code cells.

jkrick:
good catch!  That plot will only look interesting if all the files are used, but we have it setup to just download and use 1 file.  I can change the caption to reflect this.  

I also see that I still need to move the functions out to a code_src type directory.  Or would you rather leave it all as one file?

Let me know the order in which I should do these things, ie., wait for them to be on iras-tutorials and then write a PR or do a PR against the current ipac-sp-notebooks?

bsipocz:
> I also see that I still need to move the functions out to a code_src type directory. Or would you rather leave it all as one file?

We don't do code_src here, all the notebooks are expected to be self contained without outside code, or data, or requirements files.

> Let me know the order in which I should do these things, ie., wait for them to be on iras-tutorials and then write a PR or do a PR against the current ipac-sp-notebooks?

A PR here would be preferable.

jkrick:
ok, added issue #97 to remind me to change the figure caption once the notebook is merged here.

PR #96, opened 2025-04-11
MAINT: cleanup euclid4 and introduced a couple of astropy features

bsipocz:
This PR cleans up the code a bit to use more astropy features, e.g. enable WCS aware matplotlib plotting thus eliminating the need for some manual conversions.

It also cleans up the pandas usage, I haven't identified anything that was dataframe specific, so we can totally rely on Table/QTable features.

I have also added some tips and warnings around the astropy usage, but feel that all of these notebooks could still use some added narrative around Euclid itself.

### DIFF
--- a/tutorials/euclid_access/4_Euclid_intro_PHZ_catalog.md
+++ b/tutorials/euclid_access/4_Euclid_intro_PHZ_catalog.md
@@ -51,7 +51,7 @@ If you have questions about this notebook, please contact the [IRSA helpdesk](ht
 
 ```{code-cell} ipython3
 # Uncomment the next line to install dependencies if needed.
-# !pip install matplotlib pandas 'astropy>=5.3' 'astroquery>=0.4.10' fsspec firefly_client
+# !pip install matplotlib 'astropy>=5.3' 'astroquery>=0.4.10' fsspec firefly_client
 ```
 
 ```{code-cell} ipython3
@@ -59,15 +59,16 @@ import os
 import re
 import urllib
 
+import numpy as np
 import matplotlib.pyplot as plt
 
 from astropy.coordinates import SkyCoord
 from astropy.io import fits
 from astropy.nddata import Cutout2D
-from astropy.table import Table
+from astropy.table import QTable
 from astropy import units as u
 from astropy.utils.data import download_file
-from astropy.visualization import ImageNormalize, PercentileInterval, AsinhStretch, LogStretch
+from astropy.visualization import ImageNormalize, PercentileInterval, AsinhStretch, LogStretch, quantity_support
 from astropy.wcs import WCS
 
 from firefly_client import FireflyClient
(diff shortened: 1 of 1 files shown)


tutorials/euclid_access/4_Euclid_intro_PHZ_catalog.md
```diff
@@ -190,53 +189,47 @@ im_cutout= 5 * u.arcmin
...
-"
-adql
+adql = ("SELECT DISTINCT mer.object_id, mer.ra, mer.dec, "
+        "phz.flux_vis_unif, phz.flux_y_unif, phz.flux_j_unif, phz.flux_h_unif, "
+        "phz.phz_classification, phz.phz_median, phz.phz_90_int1, phz.phz_90_int2 "
+        f"FROM {table_mer} AS mer "
+        f"JOIN {table_phz} as phz "
+        "ON mer.object_id = phz.object_id "
+        "WHERE 1 = CONTAINS(POINT('ICRS', mer.ra, mer.dec), "
+                            f"BOX('ICRS', {ra_cutout}, {dec_cutout}, {size_cutout/np.cos(coords_cutout.dec)}, {size_cutout})) "
```
bsipocz, line 204: note that I switched to do a BOX search rather than a cone search to be consistent with the existing narrative and to have full coverage for our cutout image (e.g. there were a couple of galaxies missed in the corners of the image with the previous usage).

tutorials/euclid_access/4_Euclid_intro_PHZ_catalog.md
```diff
@@ -172,10 +171,10 @@ The _unif fluxes are: "Unified flux recomputed after correction from galactic ex
 We specify the following conditions on our search:
 - We select just the galaxies where the flux is greater than zero, to ensure the appear in all four of the Euclid MER images.
 - Select only objects in a circle (search radius selected below) around our selected RA and Dec
-- phz_classification =2 means we select only galaxies
-- Using the phz_90_int1 and phz_90_int2, we select just the galaxies where the error on the photometric redshift is less than 20%
+- `phz_classification = 2` means we select only galaxies
+- Using the `phz_90_int1` and `phz_90_int2`, we select just the galaxies where the error on the photometric redshift is less than 20%
 - Select just the galaxies between a median redshift of 1.4 and 1.6
-- We search just a 3 arcminute box around an RA and Dec
+- We search just a 5 arcminute box around an RA and Dec
```
bsipocz, line 177: Note: I made the narrative consistent with the code. We could choose to go the other way around though (and use the 3'x3' in the code)

tutorials/euclid_access/4_Euclid_intro_PHZ_catalog.md
```diff
@@ -190,53 +189,47 @@ im_cutout= 5 * u.arcmin
...
-result = Irsa.query_tap(adql)
-
+result_galaxies = Irsa.query_tap(adql).to_table()
+result_galaxies[:5]
+```
 
-## Convert table to pandas dataframe
-df_g_irsa = result.to_table().to_pandas()
+```{warning}
+Note that we use `to_table` above rather than `to_qtable`. While astropy's `QTable` is more powerful than its `Table`, as it e.g. handles the column units properly, we cannot use it here due to a known bug; it mishandles the large integer numbers in the `object_id` column and recast them as float during which process some precision is being lost.
```
bsipocz, line 220: I added a note about the bug as I already forgot about it and changed this line to use `to_qtable` which messes up the object_id.

Fix will come in the DB probably quicker than in astropy; so we can cleanup the warning and the code once those updates are out (not using `NA` as unit for the `long` columns).

PR #99, opened 2025-04-17
change text around final uncertainty exploration to account for case …

### LINKED ISSUE #97: BUG: correct last figure caption in number_density
This plot looks like a straight line if there is only one file downloaded and used, but looks more interesting and diagnostic if > 1 file is used.  The figure caption needs to reflect this.

jkrick:
…of single download file

This will close #97


bsipocz (approved):
Thanks!

PR #104, opened 2025-05-15
Euclid MER catalog CMD: Reverse y-axis and zoom in

troyraen:
In the Euclid MER catalog CMD figure, reverse the y-axis so that brighter stars are on top and zoom in to see detail a little better. The new figure looks like this (I can adjust x-and y-limits differently if wanted):

cmd-10k-stars-in-mer

### DIFF
--- a/tutorials/euclid_access/2_Euclid_intro_MER_catalog.md
+++ b/tutorials/euclid_access/2_Euclid_intro_MER_catalog.md
@@ -151,8 +151,8 @@ plt.errorbar(x, y, xerr=xerr, yerr=yerr,
 
 plt.xlabel('Y-H')
 plt.ylabel('Y')
-plt.xlim(-10, 10)
-plt.ylim(10, 35)
+plt.xlim(-5, 5)
+plt.ylim(30, 15)
 plt.title('10k Stars in MER catalog -- IRSA')
 ```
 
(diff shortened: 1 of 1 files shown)


bsipocz (approved):
Good eyes, thanks for spotting this!

bsipocz:
I don't want to bike-shed as it doesn't matter for this notebook, but is the gap for the very faint objects is real or is an artifact? Also, would the very red ones be an interesting group or they are just artefact again?

(That said, the cuts to zoom on the main locus in the notebook makes most sense)

troyraen:
Changed the limits and the new figure looks like this, and I'll leave it there for now.

cmd-10k-stars-in-mer

@bsipocz I really don't know. My naive guess is that maybe the really faint clump is white dwarfs (though not sure that they're quite in the right place for that) and the slightly brighter but red ones are pre-main sequence stars? 

I'll merge once the 2 pending checks come back.

bsipocz:
The two pending checks will fail, so no need to wait for them. Also, apparently we have #105.

troyraen:
@bsipocz I don't have permission to merge while checks are pending nor when checks fail, so actually I still can't merge even now. Can you do it?

bsipocz:
Ahh, indeed, I enabled branch protection 😊

PR #113, opened 2025-07-02
ENH: adding the spherex notebook

bsipocz:


bsipocz:
The CI issue is with the Euclid cloud notebook, but in a simple `Irsa.list_collections(servicetype='SIA', filter='euclid')`. Given the traceback here: https://output.circle-artifacts.com/output/job/e125fbfb-8ca2-4b7a-8504-35c940d74714/artifacts/0/_build/html/tutorials/cloud_access/euclid-cloud-access.html 
I think it's simply timing out, and we should just switch to do the call as an async TAP (even though it normally runs just fine).

Bottom line, I go ahead with the merge here for spherex as it has nothing to do with it.

Then there is a mix of fixes and workaround we can and should do:

- switch to async TAP in astroquery, or look into caching this query on the server side.
- if it's a fix in astroquery, update the version for the rendering job
- or comment out the code for this single cell as in fact it does nothing structurally important for the notebook.

PR #114, opened 2025-07-02
TMP: temporarily comment out list_collection calls

bsipocz:
This should fix/workaround the timeout issues we run into during the HTML builds. The reason is that astroquery is running a sync TAP query under the hood and the query is currently not cached on the server side.

We do fix this upstream in two, redundant ways:

- add caching on the server side
- change astroquery to use async TAP for `list_collections` (see https://github.com/astropy/astroquery/issues/3366)

In the meantime I remove these code cells from the notebooks so the rendered HTMLs can be updated.


bsipocz:
All green, so I go ahead with this.

PR #115, opened 2025-07-03
DOC: fixup spherex headings and correct link for pdf

bsipocz:


bsipocz:
I didn't branch this out from top of `main`, thus the unrelated build failure. Going ahead with the merge as none of the content here should any effect on the status.

PR #116, opened 2025-07-03
Enable nb_scroll_outputs from latest myst-nb

### LINKED ISSUE #69: ENH: adding in scrollers for long outputs
E.g. https://caltech-ipac.github.io/irsa-tutorials/tutorials/euclid_access/2_Euclid_intro_MER_catalog.html#choose-the-euclid-mer-table

Ideally this can be now done with a configitem rather than editing all the cells with the long output, but we may need to do that by contributing upstream.

jaladh-singhal:
~This PR is just a test for `scroll_outputs` functionality upcoming from upstream in https://github.com/executablebooks/MyST-NB/pull/683~

Fixes #69 

Test if the generated webpages have long outputs scrollable (bonus: also test dark mode style fixes that will come soon).

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -45,6 +45,7 @@ exclude_patterns = ['_build', 'notes', '.tox', '.tmp', '.pytest_cache', 'README.
 nb_execution_timeout = 1200
 nb_merge_streams = True
 nb_execution_mode = "cache"
+nb_scroll_outputs = True
 
 nb_execution_excludepatterns = []
 
--- a/site_requirements.txt
+++ b/site_requirements.txt
@@ -1,4 +1,5 @@
 sphinx
-myst-nb
+# To pick up newest scroller and dark-mode features
+myst-nb>=1.3
 sphinx-book-theme
 sphinx-copybutton


jaladh-singhal:
> For some outputs it feels to be an overkill to have the scroller, but one can easily get used to it I think as the overall result is better than it was before.

Yes I anticipated this, that's why I wanted irsa-tutorials to build as we have variety of notebook outputs to test. We can technically increase the max-height it gives to output cells, either upstream or by overriding mystnb.css with custom css here in our repo.

Related to this, I also see image/chart outputs are scrollable now - I don't think we want that, images should take full vertical length (right?). I'll refine the logic in my PR, I see several other issues now that irsa-tutorials have been built. But hey this is exactly the testing I need for this mystnb work - thanks for this awesome CI/CD!

bsipocz:
- you should be able to retrigger in circleCI, but you need to log in there and do it in circleCI. Your current access level should be enough to be able to do this, but DM me if there are troubles.

- Unfortunately something is not working correctly with the caching, it works for me locally but it does not seem to work on circleCI which is more than unfortunate. I should have at least an issue floating around about it, maybe more. So the current workaround you could do:
   - try and use a quicker repo, like navo-workshop. But a bunch of those are known to be failing with some heasarc problems, so your CI will never be fully green there.
   - add in a temporarily commit that adds a bunch of notebooks to the `nb_execution_excludepatterns` in `conf.py`. Given this PR is mostly a temporarily one that should work well.
   - if you have ideas how to improve CI that would be welcome, currently the HTML build is knowingly not great, and I don't yet see how to improve it besides fixing the caching (we are already rather modular for the actual testing, but the build cannot really be chopped up to pieces)

bsipocz:
ahh, yes, and I use firefox, and normally don't use dark mode and otherwise trying to remember to stay away from webdev as it's such a bottomless rabbit hole :)

jaladh-singhal:
> Yes, those were the weird ones. However, I can imagine of plotting a humungously big image, and for that we may still want a scroller? I don't know, I suppose one can get used to either approach.

Hmm I think I can hack some CSS for this that allows images to take full height but not beyond viewport height.


> Or, bold idea and way beyond the scope of your current PR, could have something on the left to collaps/expand in/out from the scroller, like we have inside jupyter? I really have no idea how involved it is to add such a UI element.

Yes, I so wanted to make this all along, but it's more work because CSS alone can't do it. I'll need to write some JS component that can handle this. I'm saving this for later (as you rightly said "webdev is a bottomless rabbit hole") if our users complaint much about scroll behavior.

bsipocz:
> Yes, I so wanted to make this all along, but it's more work because CSS alone can't do it. I'll need to write some JS component that can handle this. I'm saving this for later (as you rightly said "webdev is a bottomless rabbit hole") if our users complaint much about scroll behavior.

I wonder if that could be pulled out from jupyter itself? Not sure which repo that code lives in, but that is a totally relevant question for the mystmd team which has a convenient overlap with jupyter, too as hopefully they will also have some similar scroller config option.

bsipocz (approved):
Thanks!

PR #118, opened 2025-07-24
IRSA-6783: Update Euclid cloud notebook with HATS catalog

jaladh-singhal:
[Not ready for review] This PR will be ready once spectral columns are added in the Euclid Q1 HATS collection.

For the time being, this at least tests retrieving object id, classification, etc. for the target cone region using lsdb (that's super slow) and pyarrow (which is more lines of code but is very fast).

### DIFF
--- a/.binder/requirements.txt
+++ b/.binder/requirements.txt
@@ -2,7 +2,7 @@
 # undeclared optional dependencies of functionalities we rely on
 #
 tqdm
-numpy>=1.24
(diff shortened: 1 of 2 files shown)


troyraen:
Unfortunately, I don't have a timeline on when or even if the Q1 spectra will be included in a HATS product. We definitely want to, it's just a matter of priorities. Let's keep our eyes open for anyone outside of our group expressing interest.

Issue #98, opened 2025-04-17
ENH: re-work subject headings


jkrick:
https://jira.ipac.caltech.edu/browse/IRSA-7092

From a discussion when adding the Roman simulations number density tutorial, the headings might need a restructuring.

One possibility for dealing with the non-exclusivity of the headings is to put the notebooks under multiple headings.  I personally don't like this.  It clutters the page for me.  

One suggestion:

**Data Access**

On‑Premises Data (VO Protocols)
- Searching for 2MASS All‑Sky Atlas Images
- Searching for AllWISE Atlas Images
- Searching for contributed COSMOS images
- Searching for Spitzer Enhanced Imaging Products

Cloud Data
- IRSA cloud access introduction
- Analyzing cloud‑hosted AllWISE Source Catalog in Parquet format
- Strategies to Efficiently Work with NEOWISE Single‑exposure Source Table in Parquet
- Making Light Curves from NEOWISE Single‑exposure Source Table

Euclid Data
- Exploring Star Clusters in the Euclid ERO Data
- Euclid Q1: MER mosaics
- Euclid Q1: MER catalogs
- Euclid Q1: 1D spectra
- Euclid Q1: PHZ catalogs
- Euclid Q1: SPE catalogs
- Euclid Q1: cloud access

Simulated Data
- Querying CosmoDC2 Mock v1 catalogs
- Exploring on‑prem Roman simulated data (new tutorial—no VO protocols)
- Analyzing cloud‑hosted simulated Roman Time Domain Survey images
- Analyzing cloud‑hosted simulated Roman coadded images

**Data Visualization**
- Using Firefly visualization tools in Python to vet SEDs
- Using Firefly visualization tools to understand the light curves of Solar System objects
- Visualization of Simulated Data
- Using Firefly to explore OpenUniverse2024 mock Roman & Rubin images

**Useful Techniques**
- Parallelizing image convolution

jkrick:
[...]

PR #119, opened 2025-07-25
ENH: switching to JB2 infrastructure

### LINKED ISSUE #140: Reduce the number of headings shown in notebook TOCs
After #119 goes through, notebook TOCs will be showing all headings. Change that to have the same behavior as fornax-demo-notebooks: only show h1 and h2 until you scroll to a section that has lower-level headings and then that section in the TOC expands to show them.

bsipocz:
Opening this PR so we only iterate on the TOC structure once. 

TODOs blocking merge:

- [x] restructure TOC based on the google docs we have
- [x] enable execution in JB2
     - [x] conditional skips of notebooks (depending on OS, CI, etc.)
     - [x] install python dependencies
          - [x] ~create a universal make file to make sure we use the same in all CI~ (and then easily reproduce locally) -- added to tox for now
     - [ ] review rendered pages, collect and report missing features upstream
          - [ ] scrolling for outputs
          - [ ] add launch button
          - [ ] Move author and date metadata into notebook header -- needs some experimenting as not sure how it will look like in jupyterlab

jira: IRSA-7092

resolves https://github.com/Caltech-IPAC/irsa-tutorials/issues/98 closes https://github.com/Caltech-IPAC/irsa-tutorials/issues/123 resolves https://github.com/Caltech-IPAC/irsa-tutorials/issues/124 resolves #140 

Checklist from review:

- [x] Spitzer doesn't need "Spitzer Enhanced Imaging Products" with just one notebook, as in, we can remove the second layer under Spitzer and just have a single notebook for Spitzer.
-  In the upper right are a few symbols that don't work, can we either make them work, or remove them until they do work?  
	- [ ] jupyterlab -- **my understanding that the unclickable logo just shows if it's a notebook, but I'll be enabling the binder launch in the next follow-up, that may resolve this one issue, too, and if not I'll open an issue upstream
	- [x] edit link -- This is a known bug (https://github.com/jupyter-book/mystmd/issues/2234); TL;DR: we do tunnel through a proxy for the preview and use forks for PRs and that is just not supported. In the deployed version I expect the button to work (it does for fornax docs); though it doesn't work for The Turing Way. So we'll see once we merge this.
- [x]  **Interesting find; the reason was to have text before the top level heading. I moved it around and now it's just one title** WISE: NEOWISE: "Make light curves from ..." and "Strategies to Efficiently..." have two titles in the rendered versions. 
- [ ] **I don't think so. I think we need a better landing page, but having another TOC doesn't make much sense to me** Does it make sense to list the notebooks on the main index page, like we do for Fornax?
- [x] Comments in ignore_*_testing files about why each notebook is excluded would be really helpful.
- [x] **IMO, the heading titles in the notebook you linked below, as well as in a handful others should be reworked, but now one fewer heading layer is included** The notebook TOCs on the right: The Fornax ones only show h1 and h2 until you scroll to a section that has lower-level headings and then that section in the TOC expands to show them. On these IRSA ones, all heading levels are shown. Is that an easy setting to change? Personally, I like the Fornax behavior better.
- [ ] **Reported issue upstream: https://github.com/jupyter-book/mystmd/issues/2231#issuecomment-3336756609** -- This notebook’s TOC is so long that I can’t even scroll to the bottom of it. It cuts off at some point and I can’t see the last few sections. Seems like a bug in the rendering, but changing to Fornax style (above) should also fix it. https://circle.scientific-python.dev/output/job/ad753bcf-9ff8-48ca-a683-6c9888e57bfb/artifacts/0/_build/html/openuniverse2024preview-firefly
- [ ] **expected issue, we should port Jaladh's fix to the new theme, too: https://github.com/executablebooks/MyST-NB/pull/683** Some of the cell output is really long and not very informative. Is there a way to limit output cell height and make them scrollable? I think we’ve talked about this before but I don’t remember the answer.


bsipocz:
See #145

PR #128, opened 2025-08-06
Fix AllWISE catalog demo CI failure

troyraen:
Not sure what the problem in the CI runs is yet. We should change the s3 connection to be explicitly anonymous anyway, so starting with this to see if it magically fixes things. (I don't expect it to because, if this was the problem, CI should have crashed out around cell 7 and shouldn't have gotten all the way to 13.) I'll update this comment as I know more.

- Explicitly use an anonymous s3 connection


bsipocz (approved):
Thanks!

PR #147, opened 2025-10-01
Add spherex_cutouts and spherex_psf

troyraen:
- Transfer spherex_onprem_cutouts.md and spherex_psf.md from ipac-sp-notebooks.
- Rename spherex_onprem_cutouts.md -> spherex_cutouts.md.
- Cleanup.
- Add both notebooks to rendering.

### DIFF
--- a/tutorials/spherex/spherex_cutouts.md
+++ b/tutorials/spherex/spherex_cutouts.md
@@ -11,6 +11,7 @@ kernelspec:
   name: python3
 ---
 
+(spherex-cutouts)=
 # Download a collection of SPHEREx Spectral Image cutouts as a multi-extension FITS file
 
 ## 1. Learning Goals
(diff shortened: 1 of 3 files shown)


bsipocz (commented):
I've gone through the code cells again to try and spot points where we could use more astropy native approaches. Nothing is a blocker, but I feel all of them are good cleanup or teaching points.

bsipocz:
once you have a passing github buildhtml job, that label can be removed.

bsipocz:
Now looking back at those PRs, a few of these came up there already, but now as I had another look at the rendering I run into them again.

None of them are critical, so you could bump them in a follow-up issue.

troyraen:
@bsipocz this is passing now. Need an approving review in order to merge.

PR #151, opened 2025-10-03
Update lsdb version and remove universal_pathlib in requirements.txt

jaladh-singhal:
https://github.com/astronomy-commons/lsdb/issues/1047#issuecomment-3366994143 - the bug with universal_pathlib got resolved

### DIFF
--- a/.binder/requirements.txt
+++ b/.binder/requirements.txt
@@ -12,9 +12,7 @@ pyarrow>=10.0.1
 hpgeom
 pandas>=1.5.2
 dask[distributed]
-lsdb>=0.6.4
-# indirect dependency workaround until https://github.com/astronomy-commons/lsdb/issues/1047 is resolved and picked up as out minimum version
-universal_pathlib<0.3
+lsdb>=0.6.6
 psutil
 ray
 s3fs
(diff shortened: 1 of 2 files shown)


troyraen (commented):
Also remove it from the `pip install` cell in the notebook.

jaladh-singhal:
> Also remove it from the `pip install` cell in the notebook.

Jinx, just did!

jaladh-singhal:
I re-ran buildhtml and build-docs pipeliens, but they are still failing because of some unrelated infra error:
```
py312-buildhtml: commands[9]> bash -c 'if grep -q '"'"'Traceback .most recent call last.'"'"' /tmp/mystbuild.log; then exit 1; fi'
py312-buildhtml: commands[10]> bash -c 'rm _build/html/*thebe*.js'
rm: cannot remove '_build/html/*thebe*.js': No such file or directory
py312-buildhtml: exit 1 (0.01 seconds) /home/circleci/project> bash -c 'rm _build/html/*thebe*.js' pid=3085
  py312-buildhtml: FAIL code 1 (183.08=setup[75.65]+cmd[0.44,0.00,0.05,0.03,0.00,0.00,0.00,0.04,106.84,0.01,0.01] seconds)
  evaluation failed :( (183.12 seconds)
```

bsipocz (commented):
I would carefully double check what other implicit dependency changes would this jump in version mean, lsdb notorious with its hats --> nested_pandas-->pandas and pyarrow chains of requirements and aggressively pinning those.

bsipocz (approved):
OK, I've checked the chain of deps and it looks OK.

(well, I still don't like how they upper pin, but we can't do much about that, what's important here is that this doesn't bring in any changes to the minimums).

bsipocz:
Thanks @jaladh-singhal!

PR #153, opened 2025-10-06
Update spherex_qr -> spherex_qr2

troyraen:
Closes [IRSA-7323](https://jira.ipac.caltech.edu/browse/IRSA-7323)
Closes [IRSA-7363](https://jira.ipac.caltech.edu/browse/IRSA-7363)

Do not merge until SPHEREx QR2 has been released and is available in SIA.

SPHEREx QR2 is scheduled to be released on October ~~14~~16, available in SIA a day or two later. This PR updates notebooks to use it. I think that only the collections names will need to be updated but we'll need to test this once QR2 is available in SIA on irsatest, likely on October 13.

### DIFF
--- a/tutorials/spherex/spherex_intro.md
+++ b/tutorials/spherex/spherex_intro.md
@@ -65,7 +65,6 @@ from astroquery.ipac.irsa import Irsa
 from firefly_client import FireflyClient
 ```
 
-(spherex-query-sia)=
 ## 5. Search for SPHEREx Spectral Image MEFs that overlap coordinates you are interested in
 
 +++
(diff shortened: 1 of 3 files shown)


tutorials/spherex/spherex_intro.md
```diff
@@ -93,12 +93,12 @@ The IRSA SIA collections can be listed using using the ``list_collections`` meth
 The collections are documented at [SPHEREx Data Access: Application Program Interfaces (APIs)](https://caltech-ipac.github.io/spherex-archive-documentation/spherex-data-access#application-program-interfaces-apis)
 There are currently three collections available:
```
bsipocz, line 94: Are the old ones expected to go away? If they stay, they will show with list_collections, so it either needs a narrative update (as suggested) or some other indication that it's not only 3 available collections any more.
```suggestion
There are currently three collections available for the second Quick Release:
```

bsipocz (approved):
The text is likely needs to be updated, too.

bsipocz (approved):
Should be ready to go.

The traceback sometimes is incorrect as the error/unexpected behaviour occurs a few lines above, I suspect a non-tracebacking query was the culprit for that sporadic failure you saw.

PR #156, opened 2025-10-24
MAINT: post JB2 transition cleanup

bsipocz:
Cleaning up old configs and making files named consistently between the repos.

Also, thebe and binder interactivity is opted in now.

### DIFF
--- a/deployed_notebooks_manifest.in
+++ b/deployed_notebooks_manifest.in
@@ -1,7 +1,6 @@
 tutorials/cloud_access/cloud-access-intro.md
 tutorials/cloud_access/euclid-cloud-access.md
 tutorials/cosmodc2/cosmoDC2_TAP_access.md
-tutorials/euclid_access/.gitkeep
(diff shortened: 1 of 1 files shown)


troyraen:
Thanks, LGTM other than the build fail.

`/bin/bash: line 1: npm: command not found`

Not sure where that's coming from.

What do thebe and binder interactivity provide?

bsipocz:
somehow the node installing step is failing, I suggest waiting a bit and then restart CI

binder+thebe will add in page execution, more about it here: https://mystmd.org/guide/in-page-execution
(it's kind of experimental, but worked when I built it locally. I would see it as an alternative of the rocket launch options on the old, JB1 site)

bsipocz:
SO, imo, this should now go in.

troyraen (approved):
The interactivity looks cool. I'm a bit concerned about pushing experimental features into production at IRSA without some prior discussion. It affects the user experience. I'll leave this one to your discretion.

PR #157, opened 2025-10-24
Spherex psf update

afaisst:
Closes [IRSA-7402](https://jira.ipac.caltech.edu/browse/IRSA-7402)

Changes:
- account for the fact that the PSF zone numbering is 1-indexed. Added a "-1" at the end to capture correct zone
- modified text for PSF zone description
- modified distance calculation (accounting for 1-indexed PSF zone centers)


troyraen (approved):
Thanks @afaisst! LGTM and runs well on Fornax.

I note that the build-html failure is due to the fact that, as of today, SPHEREx QR1 data is no longer available through SIA. I'm updating the notebooks for QR2 in #153.

bsipocz (approved):

PR #160, opened 2025-10-27
Prevent GHA from running in forks

troyraen:
I wonder if we can prevent GHA from running in forks. At best it seems unnecessary and it may even be counterproductive as we're making multiple requests to archives for the same data at the same time.

### DIFF
--- a/.github/workflows/ci_publish.yml
+++ b/.github/workflows/ci_publish.yml
@@ -21,7 +21,7 @@ env:
 jobs:
 
   publish_html:
-    if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
+    if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' && github.repository == 'Caltech-IPAC/irsa-tutorials' }}
     name: Publish HTML
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/ci_publish.yml
```diff
@@ -21,7 +21,7 @@ env:
 jobs:
 
   publish_html:
-    if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
+    if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' && github.repository == 'Caltech-IPAC/irsa-tutorials' }}
```
bsipocz: Let me investigate this one a bit more as normally we turn off fork-based runs within the `on:` section above rather than in individual conditionals (we don't even want this to trigger a skipped job). 

(I didn't picked up on this bug as somehow jobs are not triggered on my own fork, but I see both you and Jaladh had them).

bsipocz (approved):
OK, let's have this in, too as a brute force measure, though I don't think it should be needed (and this will not-not trigger the job, just cancels it -- if we see no jobs after the other config adjustments, these changes can be reverted).

PR #162, opened 2025-10-30
MAINT: enabling linkchecker at built time

bsipocz:
This would do linkchecking amongst other thing to ensure all is as expected.

The PR includes some fixes and workarounds to the currently broken links and DOIs

Issue #127, opened 2025-08-05
ENH: auto-open issues about failing cron jobs


bsipocz:
This is just a reminder issue that we could definitely use this feature.

troyraen:
This is very related to https://github.com/IPAC-SW/ipac-sp-notebooks/issues/118

troyraen:
We'll create a basic GHA job that opens an issue when there's a failing cron job so that at least we'll be alerted to it. The will close this issue. The bigger task to find a proper solution will be tracked in https://github.com/IPAC-SW/irsa-datascience/issues/118.

PR #166, opened 2025-11-01
Install Node.js and mystmd for GHA

troyraen:
I think our GHA Publish HTML job is going to fail due to https://github.com/Caltech-IPAC/irsa-tutorials/pull/162#discussion_r2482279139 until we install mystmd. This is my attempt.

I'm also attempting to add a step to the Publish HTML job that will open an issue if it fails or comment on an existing issue if one exists from a previous try. We've had failures of this job both here and in fornax-demo-notebooks that we haven't noticed right away, so this would help. Related to #127 but not added to the tests job yet.

I'm saying "attempt" because I don't have much experience managing workflows, so might be doing weird things. Definitely needs a close review before merging.

### DIFF
--- a/.github/workflows/ci_publish.yml
+++ b/.github/workflows/ci_publish.yml
@@ -53,6 +53,58 @@ jobs:
           publish_dir: ./_build/html/
           commit_message: ${{ github.event.head_commit.message }}
 
+      - name: Create issue or comment if run fails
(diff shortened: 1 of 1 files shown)


troyraen:
Ok, I'll close this and open an issue. Nothing will publish to the website until it's done.

Issue #173, opened 2025-11-10
BUG: 'FireflyClient' object has no attribute 'show_fits_image'


bsipocz:
I'm not sure why this started popping now, but nevertheless, here is the tracelog (I suspect maybe something else happens before these lines that doesn't produce an error so an unrelated cell will fail instead). Seeing it on the oldestdeps run, and persistent enough to produce the failure after restart in 2 hours.

https://github.com/Caltech-IPAC/irsa-tutorials/actions/runs/19221240496/job/54939630546#step:5:812

This may also very well be an upstream issue as we could pull in some newer versions for the unpinned indirect dependencies.

cc @jaladh-singhal 

```
________________ tutorials/spherex/spherex_intro.ipynb::Cell 9 _________________
Notebook cell execution failed
Cell 9: Cell execution caused an exception

Input:
fc.show_fits_image(file_input=spectral_image_url,
             plot_id="spectral_image",
             Title="Spectral Image"
             )

Traceback:

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[1], line 1
----> 1 fc.show_fits_image(file_input=spectral_image_url,
      2              plot_id="spectral_image",
      3              Title="Spectral Image"
      4              )

AttributeError: 'FireflyClient' object has no attribute 'show_fits_image'

```

PR #177, opened 2025-11-14
BUG: fix syntax for CI to ignore unchanged notebooks

bsipocz:
#173 has unearthed that there is a bug in the scripting we have for ignoring the unchanged notebooks from testing; currently we ignore all  of them.

- this PR expects to have a failure in the oldestdeps job -- merge it only once that failure, but only that one failure is around, but after a rebase is made to remove the TEMP commit that dummy changes a notebook


bsipocz:
I'm pushing another commit to double check if maybe the logic is broken when there are multiple changes. If it still works as expected I postpone the investigation for next time when I'm not just between talks and sessions but can have full attention (and well working internet) to debug the logs.

PR #189, opened 2025-12-05
Add project board workflow

troyraen:
This workflow will trigger when Issues and PRs are opened or reopened. It will add them to the project board https://github.com/orgs/IPAC-SW/projects/5 and set the following default values:

- **Status**: "Triage" for Issues; "In Progress" for PRs
    - I'd like to have issues added to Triage for awhile, but we can reevaluate in a couple months. We might find good enough default values for other fields that we end up wanting to get rid of Triage and just have them go straight to To Do.
- **Priority**: "None"
    - The complete set of Priority options configured on the board is: "High", "Medium", and "None". 
- **Work Area**: "Content" if the item has the `content` label; "Infrastructure" if labeled with `infrastructure`.
    - If both labels are present, infrastructure takes precedence. No deep reason for this, just needed to pick one.

Once items are on the board we can / should change any field value, as appropriate.

### DIFF
--- a/.github/workflows/irsa-datasci-project-board.yml
+++ b/.github/workflows/irsa-datasci-project-board.yml
@@ -1,0 +1,68 @@
+name: Add Issues and PRs to IRSA DataSci Project Board
+
+on:
+  issues:
+    types: [opened, reopened]
+  pull_request:
+    types: [opened, reopened]
+
+jobs:
+  add-to-project:
+    runs-on: ubuntu-latest
+    env:
+      GH_TOKEN: ${{ secrets.IRSA_DATASCI_PROJECT_BOARD_TOKEN }}
+      PROJECT_ID: PVT_kwDOAnxacs4BJ3uX  # IRSA DataSci project https://github.com/orgs/IPAC-SW/projects/5
+      # Status Field
+      STATUS_FIELD_ID: PVTSSF_lADOAnxacs4BJ3uXzg56BHQ
+      TRIAGE_STATUS_ID: f971fb55
+      IN_PROGRESS_STATUS_ID: 47fc9ee4
+      # Priority Field
+      PRIORITY_FIELD_ID: PVTSSF_lADOAnxacs4BJ3uXzg56BH0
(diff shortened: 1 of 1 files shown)


troyraen:
Yes we'll want this in multiple repos. Global config sounds good. I assume we can specify which repos we want rather than getting all repos under the org? (I don't see that in the jupyter-book file but am guessing it's possible).

bsipocz:
>  I assume we can specify which repos we want rather than getting all repos under the org? (I don't see that in the jupyter-book file but am guessing it's possible).

I suppose that depends on the action you are using. The jupyter one specializes in collecting all the PRs from all the repos into one project board to help triaging while adding some extra metric and sorting; we need something different with our issue board.

jkrick (commented):
I don't know how to read this yml, so a technical review is beyond my scope.  

I do agree with the overall plan.

troyraen:
I looked in to the global .github repo option and see some pros and cons. The biggest one that sticks out to me is that a workflow added there can't be triggered by events in individual repos, it would have to run on a schedule instead. So new issues and PRs wouldn't appear on the project board immediately. I think that would cause us headaches. 

I see another option which is to make this a reusable workflow. I think it basically looks like:
- In this yaml file, change the trigger to `workflow_call`.
- In all repos where we want this to run, add a small yaml file that just calls this one.

There would still be a yaml and a secret in each repo, but all the logic defining the workflow steps would be maintained centrally in one yaml file in this repo (or a different repo if we choose, but I think this one makes sense).

What are your thoughts about that, @bsipocz ?

troyraen:
The board is meant to be the place where anyone/everyone on our team can get a handle on what's going on. The most recently opened issues are often the most important ones to see. Having a delay would make the board less useful. For me personally, there are many times when I open an issue and want to organize it on the board right then. For example, to set the priority. It would be a pain to have to wait an hour or a day and then come back to it.

bsipocz:
Actually, I don't think we need the action at all. Let's follow-up  on slack/zoom

PR #192, opened 2025-12-09
hid the functions as drop downs

jkrick:
This will close the issue number 102 over in ipac-sp-notebooks. (closes https://github.com/IPAC-SW/ipac-sp-notebooks/issues/102)

That issue was to move the functions out of the notebook into a code_src type directory as we do for fornax-demo-notebooks, but I feel it is better to hide them in drop down menus.  I also believe the ESRR reviewers mentioned wanting to see the functions, but we also don't want to clutter the notebooks.  I feel hiding them in drop downs is a happy medium.

### DIFF
--- a/tutorials/roman_simulations/roman_hlss_number_density.md
+++ b/tutorials/roman_simulations/roman_hlss_number_density.md
@@ -7,7 +7,7 @@ jupytext:
     format_version: 0.13
     jupytext_version: 1.18.1
 kernelspec:
-  display_name: python3
+  display_name: Python 3 (ipykernel)
   language: python
   name: python3
 ---


jkrick:
After discussion on slack, we have decided to hide functions in this tutorial with import statements.  This allows the functions to be easily found by people wanting to see them, and also to be lifted out of the notebooks to be run individually or in new cases by users.  I don't expect the functions in this tutorial to be used by any other tutorial, so no need here for a central directory.

bsipocz:
> After discussion on slack, we have decided to hide functions in this tutorial with import statements. This allows the functions to be easily found by people wanting to see them, and also to be lifted out of the notebooks to be run individually or in new cases by users. I don't expect the functions in this tutorial to be used by any other tutorial, so no need here for a central directory.

Right now we do not test these functions individually, so it's not guaranteed that they are standalone runnable and including all the necessary imports. This is all fine, just wanted to make this as a note. (I think it maybe possible to reset the scope for these for pytest, but I have not seen that being done for notebooks; let alone to keep multiple scopes jumping through each other, so it may or may not work).

tutorials/roman_simulations/roman_hlss_number_density.md
```diff
@@ -5,11 +5,11 @@ jupytext:
...
     format_version: 0.13
-    jupytext_version: 1.15.2
+    jupytext_version: 1.18.1
 kernelspec:
-  display_name: Python 3 (ipykernel)
-  language: python
   name: python3
+  display_name: python3
+  language: python
 ---
```
bsipocz: The order of these should not change
```suggestion
  display_name: python3
  language: python
  name: python3
---
```

bsipocz:
The only remaining CI failure was due to ads is being down and thus we have link resolution issues. So I go ahead and merge this.

Thanks @jkrick!

PR #193, opened 2025-12-10
Add Euclid HATS Introduction tutorial

troyraen:
Adds the Introduction tutorial for the Euclid Q1 Merged Objects HATS Catalog.

Adds 1 (out of 6) notebooks for #191.

PR #206, opened 2025-12-18
Add Euclid landing pages and flatten TOC

troyraen:
This is a draft of what I understood in our meeting this afternoon. I decided to do it separately from #193 to make things easier to review. This contains dummy content for the HATS notebooks so we can see what it will look like, but I'll remove that before merging.

The Q1 landing page currently contains lists of notebooks because I don't know how to make a gallery or something else that's prettier. But whenever we figure that out, hopefully the text in the list items can be used without much change.

Questions I have and problems I'm seeing:

- [x] ~~The main Euclid landing page is very short. Is that what we wanted?~~ (merged the Q1 landing page in, so now there's just one)
- [x] Is the ~~Q1~~ Euclid landing page too long? If so, how to shorten it?
- [ ] At the very bottom of all of the Euclid pages, the buttons pointing to the previous and next page all say 'SPHEREx' instead of 'Euclid' (screenshot). I suspect this is happening because I changed Euclid `title` -> `file` in toc.yml in order to add the landing pages. I tried using both `title` and `file` but that didn't make any difference. Those files have `short_title` frontmatter but it's not being used for this. Any ideas @bsipocz?
  Screenshot 2025-12-17 at 8 55 19 PM

To do before merging:

- [x] Remove the two temp commits
~- [ ] Remove the label `html rendering / skip testing`. (maybe this isn't necessary?)~ [edit by @bsipocz] - keep the label, there won't be content change in here so no need for pytesting
- [ ] Maybe make landing pages and flatten TOC for the other missions/sections as well?

### DIFF
--- a/tutorials/cloud_access/euclid-cloud-access.md
+++ b/tutorials/cloud_access/euclid-cloud-access.md
@@ -12,7 +12,6 @@ kernelspec:
   name: python3
 ---
 
-(euclid-q1-cloud)=
 # Euclid Q1: Cloud Access
 
 +++
--- a/tutorials/euclid_access/euclid.md
+++ b/tutorials/euclid_access/euclid.md
@@ -12,21 +12,21 @@ Data products include MERged mosaics of calibrated and stacked frames; combined 
 
 ## Images
 
-- [MER Mosaics](#euclid-q1-vo-mer-mosaics) — Retrieve both a full MER mosaic image and multi-wavelength cutouts, then subtract the background from the cutouts and extract sources.
+- [MER Mosaics](1_Euclid_intro_MER_images.md) — Retrieve both a full MER mosaic image and multi-wavelength cutouts, then subtract the background from the cutouts and extract sources.
 
 ## Spectra
 
-- [SIR 1D Spectra](#euclid-q1-vo-sir-1d-spectra) — Load a galaxy spectrum and plot it. Understand the wavelength, flux, and mask values.
+- [SIR 1D Spectra](3_Euclid_intro_1D_spectra.md) — Load a galaxy spectrum and plot it. Understand the wavelength, flux, and mask values.
 
 ## Catalogs
 
-- [MER Catalogs](#euclid-q1-vo-mer-catalogs) — Explore the columns in the MER final catalog, query for stars, and create a color-magnitude diagram.
-- [PHZ Catalogs](#euclid-q1-vo-phz-catalogs) — Join the PHZ and MER catalogs and do a box search for galaxies with quality redshifts, load a MER mosaic cutout of the box, and plot the cutout with the catalog results overlaid.
+- [MER Catalogs](2_Euclid_intro_MER_catalog.md) — Explore the columns in the MER final catalog, query for stars, and create a color-magnitude diagram.
+- [PHZ Catalogs](4_Euclid_intro_PHZ_catalog.md) — Join the PHZ and MER catalogs and do a box search for galaxies with quality redshifts, load a MER mosaic cutout of the box, and plot the cutout with the catalog results overlaid.
   Then plot the SIR spectrum of the brightest galaxy and look at a MER mosaic cutout of the galaxy in Firefly.
-- [SPE Catalogs](#euclid-q1-vo-spe-catalogs) — Join the SPE and MER catalogs and query for galaxies with H-alpha line detections, then plot the SIR spectrum of a galaxy with a high SNR H-alpha line measurement.
+- [SPE Catalogs](5_Euclid_intro_SPE_catalog.md) — Join the SPE and MER catalogs and query for galaxies with H-alpha line detections, then plot the SIR spectrum of a galaxy with a high SNR H-alpha line measurement.
 
 ## Special Topics
 
-- [Cloud Access](#euclid-q1-cloud) — Browse the on-cloud copy of Q1, then efficiently retrieve a MER mosaic cutout and a SIR spectrum.
-- [Early Release Observations (ERO) - Star Clusters](#euclid-ero-star-clusters) (deprecated; ERO is superseded by Q1) — Create multi-wavelength ERO image cutouts of a globular cluster, extract sources, and measure photometry.
+- [Cloud Access](../cloud_access/euclid-cloud-access.md) — Browse the on-cloud copy of Q1, then efficiently retrieve a MER mosaic cutout and a SIR spectrum.
+- [Early Release Observations (ERO) - Star Clusters](Euclid_ERO.md) (deprecated; ERO is superseded by Q1) — Create multi-wavelength ERO image cutouts of a globular cluster, extract sources, and measure photometry.
   Then load Gaia sources at the location of the globular cluster, match with Euclid ERO catalogs, and visualize the results with Firefly.
--- a/tutorials/euclid_access/1_Euclid_intro_MER_images.md
+++ b/tutorials/euclid_access/1_Euclid_intro_MER_images.md
@@ -12,7 +12,6 @@ kernelspec:
   name: python3
 ---
 
-(euclid-q1-vo-mer-mosaics)=
 # Euclid Q1: MER Mosaics
 
 +++
--- a/tutorials/euclid_access/2_Euclid_intro_MER_catalog.md
+++ b/tutorials/euclid_access/2_Euclid_intro_MER_catalog.md
@@ -12,7 +12,6 @@ kernelspec:
   name: python3
 ---
 
-(euclid-q1-vo-mer-catalogs)=
 # Euclid Q1: MER Catalogs
 
 +++
--- a/tutorials/euclid_access/3_Euclid_intro_1D_spectra.md
+++ b/tutorials/euclid_access/3_Euclid_intro_1D_spectra.md
@@ -12,7 +12,6 @@ kernelspec:
   name: python3
 ---
 
-(euclid-q1-vo-sir-1d-spectra)=
 # Euclid Q1: SIR 1D Spectra
 
 +++
--- a/tutorials/euclid_access/4_Euclid_intro_PHZ_catalog.md
+++ b/tutorials/euclid_access/4_Euclid_intro_PHZ_catalog.md
@@ -12,7 +12,6 @@ kernelspec:
   name: python3
 ---
 
-(euclid-q1-vo-phz-catalogs)=
 # Euclid Q1: PHZ Catalogs
 
 +++
--- a/tutorials/euclid_access/5_Euclid_intro_SPE_catalog.md
+++ b/tutorials/euclid_access/5_Euclid_intro_SPE_catalog.md
@@ -12,7 +12,6 @@ kernelspec:
   language: python
 ---
 
-(euclid-q1-vo-spe-catalogs)=
 # Euclid Q1: SPE Catalogs
 
 +++
--- a/tutorials/euclid_access/Euclid_ERO.md
+++ b/tutorials/euclid_access/Euclid_ERO.md
@@ -13,7 +13,6 @@ kernelspec:
   name: python3
 ---
 
-(euclid-ero-star-clusters)=
 # Exploring Star Clusters in the Euclid ERO Data
 
 


tutorials/euclid_access/euclid-q1.md
```diff
@@ -0,0 +1,47 @@
...
+
+IRSA also serves the Euclid Q1 data in FITS and Apache Parquet file formats from an Amazon Web Services (AWS) S3 cloud storage bucket.
+This tutorial demonstrates access using the Python libraries [S3Fs](https://s3fs.readthedocs.io/en/latest/) and [Astropy](https://astropy.readthedocs.io/en/stable/).
+S3 is well-suited for most use cases and is especially efficient for large-scale data access.
+
+- [Cloud Access](#euclid-q1-cloud) — Browse the S3 bucket, then efficiently retrieve a MER mosaic cutout and a SIR spectrum.
+- See [Merged Objects HATS Catalog](#euclid-q1-hats-series) for efficient access to the on-cloud catalogs.
+
+(euclid-q1-hats-series)=
+## Merged Objects HATS Catalog
```
bsipocz: maybe have it one level down, given these catalogs are cloud ones anyway?
```suggestion
### Merged Objects HATS Catalog
```

bsipocz:
>  The main Euclid landing page is very short. Is that what we wanted?
Is the Q1 landing page too long? If so, how to shorten it?

Merge them together, so Q1 will be a section with a heading link on the Euclid landing page. So it will be possible to link to it, but it flattens the structure a little bit (and we can still have nested toc on the side.

bsipocz:
> Remove the label html rendering / skip testing. (maybe this isn't necessary?)

Since this PR won't change the notebook content, it can stay on

troyraen:
> Merge them together

Done, but let me know if it is as you envisioned.

tutorials/euclid_access/euclid.md
```diff
@@ -0,0 +1,61 @@
...
+
+IRSA also serves the Euclid Q1 data in FITS and Apache Parquet file formats from an Amazon Web Services (AWS) S3 cloud storage bucket.
+This tutorial demonstrates access using the Python libraries [S3Fs](https://s3fs.readthedocs.io/en/latest/) and [Astropy](https://astropy.readthedocs.io/en/stable/).
+S3 is well-suited for most use cases and is especially efficient for large-scale data access.
+
+- [Cloud Access](#euclid-q1-cloud) — Browse the S3 bucket, then efficiently retrieve a MER mosaic cutout and a SIR spectrum.
+- See [Merged Objects HATS Catalog](#euclid-q1-hats-series) for efficient access to the on-cloud catalogs.
+
+(euclid-q1-hats-series)=
+### Merged Objects HATS Catalog
```
troyraen: I keep wondering about these anchors. They're great because we can change file names and heading text without breaking the links, and that's super helpful. But on the other hand, they make the links totally useless outside of the website. When looking at the markdown files, even on github where the markdown is rendered, the links don't work at all and the reader is lost. I don't know what to do about that.

bsipocz: hmm, that's a good point, I totally was focusing on the rendered outputs and was not even thinking about the notebooks themselves. 

If we don't yet track this issue upstream, let me add it to the "papercut with notebooks" list.

bsipocz (commented):
apparently I started a review, but now I cannot leave single comments, so just release the whole review as is

tutorials/cloud_access/euclid-cloud-access.md
```diff
@@ -11,7 +12,8 @@ kernelspec:
   name: python3
 ---
 
-# Euclid Q1: cloud access
+(euclid-q1-cloud)=
```
jaladh-singhal: Ooh custom anchor links - that's great!

tutorials/euclid_access/euclid.md
jaladh-singhal: That's a valid complaint but isn't that expected?! I mean a renderer decides how to render anchor - this syntax doesn't mean anything to Github markdown same as how frontmatter doesn't mean anything. So Github markdown just renders it as text and rather creates heading-text-based anchors. Whereas it has a meaning for the JB website renderer so it shows up as expected. 

Do you expect people to linking anchors or other items from the Github markdown doc as opposed to from our documentation website?

tutorials/euclid_access/euclid.md
```diff
@@ -0,0 +1,61 @@
...
+
+The Euclid Q1 Merged Objects HATS Catalog comprises fourteen of the original Euclid Q1 catalogs, joined on the Euclid object ID, into a single dataset.
+The catalog is in Apache Parquet file format and is stored on AWS S3.
+These tutorials use the Python [Apache PyArrow](https://arrow.apache.org/docs/python/index.html) library for efficient data access.
+HATS, Parquet, and PyArrow are described further on [IRSA Parquet Catalogs](https://irsa.ipac.caltech.edu/docs/parquet_catalogs/).
+These methods are well-suited for most use cases and are especially good for large-scale analyses.
+
+- [Introduction](#euclid-q1-hats-intro) — Understand the content and format of the Euclid Q1 Merged Objects HATS Catalog, then perform a basic query.
```
jaladh-singhal: is there a specific reason for linking the custom anchors for H1 heading for a notebook rather than linking the notebook document path itself (assuming that is possible)? 


```suggestion
- [Introduction](../parquet-catalog-demos/euclid-q1-hats/1-euclid-q1-hats-intro) — Understand the content and format of the Euclid Q1 Merged Objects HATS Catalog, then perform a basic query.
```

I'm not certain but this may resolve several complaints with Github markdown rendering, because Github recognizes path-based linking IIRC.

toc.yml
```diff
@@ -9,18 +9,14 @@ project:
...
-        - title: Euclid Quick Release 1
-          children:
-            - file: tutorials/euclid_access/1_Euclid_intro_MER_images.md
-            - file: tutorials/euclid_access/2_Euclid_intro_MER_catalog.md
-            - file: tutorials/euclid_access/3_Euclid_intro_1D_spectra.md
-            - file: tutorials/euclid_access/4_Euclid_intro_PHZ_catalog.md
-            - file: tutorials/euclid_access/5_Euclid_intro_SPE_catalog.md
+            - pattern: tutorials/euclid_access/*_Euclid_intro_*.md
             - file: tutorials/cloud_access/euclid-cloud-access.md
+            - pattern: tutorials/parquet-catalog-demos/euclid-q1-hats/*-euclid-q1-hats-*.md
```
jaladh-singhal: This flatter TOC renders nicely - thanks!

tutorials/euclid_access/euclid.md
jaladh-singhal: Thinking more about it...I think we can get best of both worlds by refraining from using these special header targets for H1 headings because then you can just link the document path.

tutorials/euclid_access/euclid.md
bsipocz: it may not create hovers though and I suspect neither of these would work in a jupyterlab setting?

tutorials/euclid_access/euclid.md
```diff
@@ -0,0 +1,61 @@
...
+
+This notebook describes the [Euclid ERO](https://irsa.ipac.caltech.edu/data/Euclid/docs/overview_ero.html) data products and demonstrates access.
+
+- [Star Clusters](#euclid-ero-star-clusters) — Create multi-wavelength ERO image cutouts of a globular cluster, extract sources, and measure photometry.
+  Then load Gaia sources at the location of the globular cluster, match with Euclid ERO catalogs, and visualize the results with Firefly.
+
+(euclid-q1)=
+## Quick Release 1 (Q1)
+
+These notebooks describe the [Euclid Q1](https://irsa.ipac.caltech.edu/data/Euclid/docs/overview_q1.html) data products and demonstrate access using various methods.
```
jkrick:
```suggestion
These notebooks describe the [Euclid Q1](https://irsa.ipac.caltech.edu/data/Euclid/docs/overview_q1.html) data products and demonstrate access using two different techniques; what we will call VO and HATS below.```

**Review by jkrick — COMMENTED** [review]
General comment: I don't think the organization is clear, specifically how we differentiate between the currently divided sets of notebooks.  In the drop downs on the left column they are differentiated by "VO" and "HATS", with one lone "cloud access".  

These two words "VO" and "HATS" mean nothing to users, so I would suggest we brainstorm other ideas.  Is the main difference between them the end database?  ie/, one is in parquet format and one is not?  So maybe the main difference that is a valuable distinction to users is that one does large scale searches faster, and one is more appropriate for small scale searches?  Looking at this from a user perspective, we need to be able to answer why we are showing users more than one way to do the same task.  

Maybe this is heretical, but maybe the merged catalog gets its own level because it is its own data product.  So I am suggesting we have Euclid as the main landing page.  That subdivides into ERO, Q1, and Merged Objects Catalog (Q1). We'll figure out what to do with Q2 when that hits in 6 months and we decide to reorganize this anyway.   Yes, I like this.  Then recommend removing "HATS" and "VO" from the front of the short names of the notebooks.

In my suggested scheme, Jaladh's "cloud access" would go under Q1

Probably this means my individual comments below aren't relevant anymore because the overall structure will change, so I won't go back and mess with those if you do decided to adopt this change.

File: tutorials/euclid_access/euclid.md (hunk 2 above)
Comment (bsipocz) [reply tutorials/euclid_access/euclid.md]: Just to rephrase, if we need to drop one of the 3 access points, I would totally drop the Github rendering. It doesn't matter how things look on github as our end user products are the website and the notebooks. (and we know a bunch of myst stuff is already broken in github markdown rendering as it doesn't speak the myst flavour of the format)

**Comment by vandesai1** [comment]
To answer your specific question, I do think that the Euclid landing page is too long. I think it will help to have the gallery, because it will be easier to pick out which text is a notebook and which text isn't.

File: tutorials/euclid_access/euclid.md (hunk 2 above)
Comment (troyraen) [reply tutorials/euclid_access/euclid.md]: The github rendering is just one example of how users interact with the markdown files, which we definitely expect them to do. I don't think it's a big deal that the frontmatter and admonitions look as they do, but links not working is significant. Not saying it's more important than the other 2 access points, just that there is tension here that is not insignificant.

Are there good reasons not to use standard markdown syntax to link to sections? For example, when I want to put a link in the euclid hats notebooks to the hats section of this landing page, standard markdown syntax would look something like `[link name](../euclid_access/euclid.md#merged-objects-hats-catalog)` and no anchor (or myst label or header target) would be needed. That will be more work for us to maintain, but if it means the links will work everywhere then we should do it.

File: tutorials/euclid_access/euclid-q1.md [hunk 7]
```diff
@@ -0,0 +1,47 @@
...
+
+(euclid-q1-cloud-series)=
+## Cloud Access to Images, Spectra, and Catalogs
+
+IRSA also serves the Euclid Q1 data in FITS and Apache Parquet file formats from an Amazon Web Services (AWS) S3 cloud storage bucket.
+This tutorial demonstrates access using the Python libraries [S3Fs](https://s3fs.readthedocs.io/en/latest/) and [Astropy](https://astropy.readthedocs.io/en/stable/).
+S3 is well-suited for most use cases and is especially efficient for large-scale data access.
+
+- [Cloud Access](#euclid-q1-cloud) — Browse the S3 bucket, then efficiently retrieve a MER mosaic cutout and a SIR spectrum.
+- See [Merged Objects HATS Catalog](#euclid-q1-hats-series) for efficient access to the on-cloud catalogs.
```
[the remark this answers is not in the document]
troyraen: This section is incomplete without covering catalogs, and HATS is the answer there - "cloud access" and "hats" is just not a clean split. I wondered about not having a separate section for hats but it seems like there's enough to say that it warrants its own section. Maybe that's wrong? Maybe your suggestion below to make hats a subsection of cloud access is the right answer. It at least makes more sense than the current organization.

vandesai1:
I've got a question out to ENSCI that will help with the organization, I think. I'm not sure if Q1 will be superseded by DR1, and if DR1 will be superseded by DR2. If they are superseded, then we will not need *new* notebooks. We will just update the existing notebooks.

Issue #233, opened 2026-01-30
Auto-generate gallery style cards on index pages


jaladh-singhal:
After #219 and #220 are merged and when we have short summaries of each notebook in yaml, the grid of cards (notebook gallery) can be abstracted as a function-like entity. Such entity will allow us to reuse our gallery element in any index page we want just by passing a list of notebook files. And the style and layout of gallery element can be controlled at one place to affect all uses of it.

Ideas to implement it:
- https://github.com/jupyter-book/jupyter-book/blob/main/docs/src/create_gallery.py to produce https://jupyterbook.org/stable/gallery/
- custom local plugin: https://mystmd.org/guide/plugins-dry (I prefer this more!)

PR #236, opened 2026-02-03
Add yaml file with notebook metadata

troyraen:
This adds a yaml file with notebook names, paths, and descriptions. Currently, the names and descriptions are copies of those on the landing pages - #233 will switch the landing pages to use this file and remove the duplication. #237 will use this yaml to populate the descriptions for the Fornax index page.

This also adds irsa-hats-with-lsdb.md to deploy_to_fornax_manifest.in. I'm guessing leaving it out was just an oversight.

### DIFF
--- a/notebook_metadata.yml
+++ b/notebook_metadata.yml
@@ -1,87 +1,87 @@
-- name: Cloud Access Intro
-  path: tutorials/cloud_access/cloud-access-intro.md
+- title: Cloud Access Intro
+  file: tutorials/cloud_access/cloud-access-intro.md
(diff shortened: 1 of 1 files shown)


notebook_metadata.yml
```diff
@@ -0,0 +1,87 @@
+- name: Cloud Access Intro
+  path: tutorials/cloud_access/cloud-access-intro.md
```
bsipocz: Not a blocker for now, but for consistency with TOC naming, I would recommend using "title" instead of name and "file" instead of path

PR #241, opened 2026-02-05
Enable execution on CircleCI for more notebooks

### LINKED ISSUE #209: Add more notebooks to CircleCI execution
@troyraen - after this is in, we can look into enabling back the execution of more of the notebooks on circleCI, using the assumption that if only a subset of notebooks are executed they would fare better with resources.

_Originally posted by @bsipocz in https://github.com/Caltech-IPAC/irsa-tutorials/issues/203#issuecomment-3663665837_

bsipocz:
This requires circleCI specific testing via an SSH session as well as disabling a lot of the test skip logic, leaving as draft while iterating on it.

This closes #209

### DIFF
--- a/ignore_tutorials/ignore_circleci_testing
+++ b/ignore_tutorials/ignore_circleci_testing
@@ -1,13 +1,8 @@
-# ignore these 10 now as excluding execution for these is a combination that passes JB2 building on CircleCI.
-# Note that the first 2 are excluded for all kinds of testing; while the parallelize causes issues for GHA JB2 builds and non linux testing, too.
+# ignore these as they run out of resources on CircleCI.
+# Note that we don't duplicate the listing of notebooks that are ignored for all types of testing.
 #
-tutorials/euclid/euclid-cloud-access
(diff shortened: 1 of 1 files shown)


bsipocz:
I was playing some significant amount of time today to find a combination that passes. At some point the following seemed to be a good combo, with the cloud access intro one either added or not, but then I couldn't reproduce the passing status. Looking at the resource log it seems that there wasn't really proper garbage collection during the session as the restarts were not real restarts. 
So, apparently one needs to do this with brand new sessions and that just makes the task even more tedious. 

So, I could imagine a better solution would be to enable back everything, and we keep an eye out to the resource problem in individual PRs -- then it's more clear which ones are affected (as we only test the ones that have been changed).

```
tutorials/roman_simulations/roman_hlss_number_density
tutorials/openuniversesims/openuniverse2024_roman_simulated_timedomainsurvey
tutorials/cloud_access/cloud-access-intro
tutorials/cloud_access/euclid-cloud-access
```

PR #256, opened 2026-02-17
Adding licence to the rendered pages

bsipocz:


troyraen (commented):
Thanks! I see that this adds an icon that links to the generic license. Is there a way to link to the license in the repo? The one in the repo has the copyright holder filled in.

bsipocz:
I'm not sure why that is done by default rather than linking the actual license file. IMO it should be OK to leave that to the users to figure out, but we can also instead just hard-wire the URL.

PR #258, opened 2026-02-19
Add SPHEREx Source Discovery Tool

troyraen:
Closes IPAC-SW/ipac-sp-notebooks/issues/158

The notebook and code have previously been reviewed.

I made the following changes for the deployment:

- Convert to myst markdown. Standardize headings. Fix up Jupyter and Fornax terminology. Reorder list of files in README.
- Add entry to TOC and SPHEREx landing page
- Ignore when executing notebooks for rendering
- Add to Fornax deployment

Made the following changes to the content, as requested or approved by authors:

- Add a description of the Source Discovery Tool at top of notebook and README
- Add acknowledgement of Zafar Rustamkulov
- Reorder author list

### DIFF
--- a/tutorials/spherex/spherex_source_discovery/README.md
+++ b/tutorials/spherex/spherex_source_discovery/README.md
@@ -30,7 +30,7 @@ All necessary packages and tools (SExtractor) are listed in the `conda-spherex_s
 
 To create a conda environment with the dependencies on your local machine use:
   ```
-  conda env create --f conda-spherex_sdt.yml
+  conda env create --file conda-spherex_sdt.yml
   ```
 
 Then, make the environment available in the list of kernels for your JupyterLab:
--- a/tutorials/spherex/spherex_source_discovery/spherex_source_discovery_tool_demo.md
+++ b/tutorials/spherex/spherex_source_discovery/spherex_source_discovery_tool_demo.md
@@ -54,7 +54,7 @@ All necessary packages and tools (e.g., <code>SExtractor</code>) are listed in t
 
 To create a conda environment with the dependencies on your local machine use:
   ```
-  conda env create --f conda-spherex_sdt.yml
+  conda env create --file conda-spherex_sdt.yml
   ```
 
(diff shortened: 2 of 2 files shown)


bsipocz:
I would say we should skip testing on this for now, so shipping this is not blocker. There is precedent as we don't test the neowise notebooks either.

tutorials/spherex/spherex_sdt/conda-sdt_env.yml
```diff
@@ -0,0 +1,24 @@
+name: sdt_env
```
bsipocz: the env name should preferably be more verbose, or if for some reason we need to be terse at minimum be the same as either the directory or the notebook filename.

tutorials/spherex/spherex_sdt/spx_sdt/__init__.py
bsipocz: I would reconsider the naming of this directory: make it more obvious what is inside. I would definitely spell out the words in the abbreviation, after all people will be importing things and won't need to type out things all the time.

tutorials/spherex/spherex_sdt/README.md
bsipocz: rename this file, so the rendered URL will make more sense (as we flatten out the namespace it's better to have explicit name than a generic README.

bsipocz: (also, consider if it would be useful to add this to the HTMLs as a help page)

bsipocz (commented):
Some generic review comments; mostly about naming files, namespaces and the content of the environment file.

None of these are blockers, but all would make the usage experience smoother and cleaner.

tutorials/spherex/spherex_sdt/README.md
troyraen: I thought about rendering this, but the overview and setup instructions are repeated in the notebook, so the only content in here that's not also in the notebook is the directory contents. So I opted not to render this README, and instead I just pointed the TOC and the spherex landing page directly to the notebook. That seemed better than creating another level in the hierarchy for this readme / landing page. Thoughts?

PR #266, opened 2026-02-27
Move authors to frontmatter

troyraen:
This moves the notebook author lists out of the "About this Notebook" sections and into the frontmatter so that they show up under the notebook titles. Also removes the generic "IRSA scientists and developers". We need to do this for the SPHEREx source discovery notebook rather quickly but we want to do it for all notebooks anyway.

It might be good to clean up a few of these author lists but I'll bring that up separately so that it doesn't hold up this PR.

### DIFF
--- a/index.md
+++ b/index.md
@@ -8,6 +8,4 @@ They cover topics like querying IRSA, working with catalogs in Parquet format, v
 
 ## About these notebooks
 
-**Authors:** IRSA Scientists and Developers wrote and maintain these notebooks.
-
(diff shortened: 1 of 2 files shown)


bsipocz:
> I think circleci is failing because I touched all notebooks and that causes grep -vf changed_tutorials all_tutorials to return exit code 1, which is treated as a failure. So we probably need to update tox.ini either here or in a separate pr.

Oh, I thought I fixed all of those already, but now I see it's an unmerged changed in an open PR. I'll cherry-pick those changes into a separate PR.

PR #296, opened 2026-04-07
Add mention of 'include_metadata_tables'

### LINKED ISSUE #215: Explain how to list all catalogs using astroquery >=0.4.11
The default behavior of `Irsa.list_catalogs()` changed in astroquery v0.4.11 such that it no longer outputs image metadata tables, only catalogs. There is a new kwarg that can be used to get the old default behavior, `include_metadata_tables=True`. This causes a bit of trouble for our Euclid notebooks, many of which use that function to show all the Euclid tables. Trouble is that 25/26 of the Euclid tables are "metadata" tables, so users with astroquery >=0.4.11 don't see them and it's confusing. Unfortunately, `include_metadata_tables` is a new kwarg, so we can't just add that to the notebooks because it will error for users with <0.4.11. Probably the best we can do is to add a code comment to those cells telling people that if they only see one catalog output then they should add the kwarg `include_metadata_tables=True` to see all of them.

bsipocz:
This is to resolve #215 

I went ahead and updated the notebooks to default assume the users have the latest astroquery, but also added a note box that hopefully provide enough guidance for those with an older version

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -66,6 +66,9 @@ commands =
 
     buildhtml: bash -c 'cat ignore_tutorials/ignore_rendering_execution >> ignore_testing'
 
+    # We use astroquery 0.4.11+ feature in some of the notebooks, ignore them for oldestdeps testing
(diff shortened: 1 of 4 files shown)


troyraen (approved):
Thanks!

bsipocz:
Build error is unrelated, so I go ahead with the merge. I have also updated the astroquery docs to explicitly mention `include_metadata_tables`.

PR #298, opened 2026-04-15
Add `notebook-gallery` plugin to auto-generate gallery style cards on index pages

### LINKED ISSUE #233: Auto-generate gallery style cards on index pages

jaladh-singhal:
Fixes #233

Added a custom local MyST plugin (`notebook-gallery` directive) that auto-generates a notebook gallery we have for the Euclid index page and used it for all other mission pages.

### Design decisions:
@bsipocz you probably care about this more (since we had past conversations about this):

- Implemented this as a MyST JS plugin (the `.mjs` file) rather than a Python script, as the [MyST plugin system](https://mystmd.org/guide/plugins) integrates JS plugins cleanly at build time and avoids a separate generation step for [executable plugins](https://mystmd.org/guide/executable-plugins). Also if we want more complexity we can easily migrate to [widgets](https://mystmd.org/guide/widgets) from JS route - right now we don't need it because we are not working with HTML DOM directly
- Took inspiration from JB [example-js-plugin](https://github.com/jupyter-book/example-js-plugin/)
- The plugin depends on the `yaml` npm package for parsing `notebook_metadata.yml`. So a build step is required and I had to add `package.json` (+`package-lock.json`) and a bundling step via `esbuild` in `tox.ini` (as opposed to a single self-contained `.mjs` file). 
   - This has an added benefit: we can eventually move `plugins/` to a separate `irsa-myst-plugins` repo and publish it as npm/GH package if we decide to reuse it across projects or to contribute upstream.

### Things to check:
- Test mission index pages: WISE, SPHEREx, Simulated Data, Techniques & Tools, Euclid
- Title differences in the Euclid gallery cards at ops and this PR build
   - Can be fixed if we add short titles from toc.yml to notebook_metadata.yml (they should be reconciled)
   - If not, I can try modifying the plugin to override title for some notebooks but it will make code a bit more complex
- [Exception cases](https://circle.scientific-python.dev/output/job/ca66142c-b2d2-42a5-97ed-8a5b0be20798/artifacts/0/_build/html/euclid#notebook-gallery-exception-handling-tests) in `euclid.md` will be removed before merging

### DIFF
--- a/myst.yml
+++ b/myst.yml
@@ -25,7 +25,11 @@ project:
          - https://irsa.ipac.caltech.edu/cgi-bin/Gator/nph-scan?projshort=SPITZER
          # GH is rather flaky for CI right now, temporarily measure
          - 'https://github.com/**'
-         
+  plugins:
+    # note: we need to use the built plugin bundle here so make sure to build it with
+    # `cd plugins && npm install && npm run build` after any changes to the plugins/src/
+    - plugins/dist/notebook-gallery.mjs
+    
 extends:
   - toc.yml
 site:
--- a/tox.ini
+++ b/tox.ini
@@ -98,7 +98,7 @@ commands =
 
     # Using srtict so we fail with trackbacks and debug mode to have a richer log
     # For full build we disable parallel runs to easy server load
-    buildhtml: bash -c "jupyter-book build --execute --execute-parallel 1 --html -d"
+    buildhtml: bash -c "jupyter-book build --execute --execute-parallel 1 --html --strict -d"
 
 pip_pre =
     predeps: true
--- a/tox.ini
+++ b/tox.ini
@@ -93,6 +93,9 @@ commands =
 
     buildhtml: bash -c "echo 'Notebooks ignored (not tested/executed) in this job:'; cat ignore_execute | sort | uniq"
 
+    # Build the MyST plugin bundle before building the site with jupyter-book
+    buildhtml: bash -c "cd plugins && npm install && npm run build"
+
     # Using srtict so we fail with trackbacks and debug mode to have a richer log
     # For full build we disable parallel runs to easy server load
     buildhtml: bash -c "jupyter-book build --execute --execute-parallel 1 --html --strict -d"
--- a/tox.ini
+++ b/tox.ini
@@ -98,7 +98,7 @@ commands =
 
     # Using srtict so we fail with trackbacks and debug mode to have a richer log
     # For full build we disable parallel runs to easy server load
-    buildhtml: bash -c "jupyter-book build --execute --execute-parallel 1 --html --strict -d"
+    buildhtml: bash -c "jupyter-book build --execute --execute-parallel 1 --html -d"
 
 pip_pre =
     predeps: true
--- a/tutorials/euclid/euclid.md
+++ b/tutorials/euclid/euclid.md
@@ -24,18 +24,3 @@ tutorials/euclid/euclid-cloud-access.md
 tutorials/euclid/euclid_clusters_tutorial.md
 tutorials/euclid/Euclid_ERO.md
 ```
-
-## notebook-gallery exception handling tests
-### Invalid notebook path
-```{notebook-gallery} notebook_metadata.yml
-tutorials/euclid/1_Euclid_intro_MER_images.md
-tutorials/euclid/does_not_exist.md
-tutorials/euclid/2_Euclid_intro_MER_catalog.md
-```
-
-### Invalid metadata path
-```{notebook-gallery} does_not_exist.yml
-tutorials/euclid/1_Euclid_intro_MER_images.md
-tutorials/euclid/2_Euclid_intro_MER_catalog.md
-tutorials/euclid/3_Euclid_intro_1D_spectra.md
-```
--- a/tutorials/euclid/euclid.md
+++ b/tutorials/euclid/euclid.md
@@ -12,69 +12,15 @@ Data products include:
 - Catalogs of MER objects, photometric redshifts and classifications (PHZ), and spectroscopic redshifts and line measurements (SPE);
 - Merged Objects Catalog (created by IRSA) containing the MER, PHZ, and SPE catalogs in a single HATS Catalog.
 
-````{grid} 1 2 2 3
-
-```{card}
-:link: 1_Euclid_intro_MER_images.md
-:header: [MER Image Mosaics →](1_Euclid_intro_MER_images.md)
-Retrieve both a full MER mosaic image and multi-wavelength cutouts, then subtract the background from the cutouts and extract sources.
+```{notebook-gallery} notebook_metadata.yml
+tutorials/euclid/1_Euclid_intro_MER_images.md
+tutorials/euclid/3_Euclid_intro_1D_spectra.md
+tutorials/euclid/2_Euclid_intro_MER_catalog.md
+tutorials/euclid/4_Euclid_intro_PHZ_catalog.md
+tutorials/euclid/5_Euclid_intro_SPE_catalog.md
+tutorials/euclid/merged-objects-hats-catalog/1-euclid-q1-hats-intro.md
+tutorials/euclid/merged-objects-hats-catalog/4-euclid-q1-hats-magnitudes.md
+tutorials/euclid/euclid-cloud-access.md
+tutorials/euclid/euclid_clusters_tutorial.md
+tutorials/euclid/Euclid_ERO.md
 ```
-
-```{card}
-:link: 3_Euclid_intro_1D_spectra.md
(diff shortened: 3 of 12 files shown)


jaladh-singhal:
This is relevant for build step changes in tox: https://app.circleci.com/pipelines/github/Caltech-IPAC/irsa-tutorials/849/workflows/b6cd0cd5-fd7d-43d6-9d30-86bd8a2c1fc6/jobs/826?invite=true#step-103-10522_83

image

Also note L265: the plugin is loaded successfully (right now we just have 1 directive and that's what it tells us)

myst.yml
```diff
@@ -25,7 +25,11 @@ project:
          - https://irsa.ipac.caltech.edu/cgi-bin/Gator/nph-scan?projshort=SPITZER
          # GH is rather flaky for CI right now, temporarily measure
          - 'https://github.com/**'
-         
+  plugins:
+    # note: we need to use the built plugin bundle here so make sure to build it with
+    # `cd plugins && npm install && npm run build` after any changes to the plugins/src/
+    - plugins/dist/notebook-gallery.mjs
```
bsipocz, line 31: This is OK for now, but ultimately we may want to ship the plugin separately from this repo.

tox.ini
```diff
@@ -93,9 +93,12 @@ commands =
 
     buildhtml: bash -c "echo 'Notebooks ignored (not tested/executed) in this job:'; cat ignore_execute | sort | uniq"
 
+    # Build the MyST plugin bundle before building the site with jupyter-book
+    buildhtml: bash -c "cd plugins && npm install && npm run build"
```
bsipocz, line 97: It's OK here, but we will need to make sure npm is available for tox? I suppose we can kind of rely on that it's available already for JB so we can just piggy-back?

tox.ini
```diff
@@ -93,9 +93,12 @@ commands =
...
+
     # Using srtict so we fail with trackbacks and debug mode to have a richer log
     # For full build we disable parallel runs to easy server load
-    buildhtml: bash -c "jupyter-book build --execute --execute-parallel 1 --html --strict -d"
```
bsipocz, line 98: If this goes in with the merge we will need to open a follow-up issue to ensure that eventually we turn back on the checks.

tutorials/euclid/euclid.md
```diff
@@ -12,69 +12,30 @@ Data products include:
...
-```{card}
-:link: Euclid_ERO.md
-:header: [ERO Star Clusters →](Euclid_ERO.md)
-Create multi-wavelength ERO image cutouts of a globular cluster, extract sources, and measure photometry. Match Gaia sources with Euclid ERO catalogs, then visualize with Firefly.
-```
-
-````
+```{notebook-gallery} notebook_metadata.yml
+tutorials/euclid/1_Euclid_intro_MER_images.md
+tutorials/euclid/3_Euclid_intro_1D_spectra.md
```
bsipocz, line 17: I like this approach very much; thanks Jaladh. 

However; I already start seeing a list of options for improvements. What would work best, comment on the plugin itself or here argue for the usecase?

bsipocz (approved):
This looks great, thank you so much.

I think 1) this should go in as first version before we jump into with new ideas for improvements (with the removal of the test sections of course -- those could maybe go into a new test md file in the plugin itself, or added as a test md file if/when the plugin will live in its separate repo)

myst.yml
jaladh-singhal: I agree - and note the plural in "plugins", I envision it as a suite of plugins that we can maintain, ship and potentially share with MyST community.

tox.ini
jaladh-singhal: Yes, it seemed to work locally as well as on the CircleCI/GA runner had `npm` pre-installed (which is common for most of their runner images). I don't think we particularly piggybacked on JB because we install JB via pip as python package not as an npm package.

That being said, I didn't think this through and we should be aware of it if we run into npm missing issues. Another reason why plugins/ should be it's own repo and we don't maintain the build step here at irsa-tutorials.

tox.ini
jaladh-singhal: I will revert it before merging. 

The first build failed (when strict was on) not because of plugin installation step  but because of some notebook or something (I couldn't tell because the logs are way more verbose than my attention span). But yeah, I'm afraid it's becoming more common to remove strict for these kind of PR builds, maybe we should make it a label like skip testing GH label?

tutorials/euclid/euclid.md
jaladh-singhal: I mean it's upto you. If you wish to get this merged, then you can open an issue listing all improvements you see. Else you can comment here and I can iterate within the PR

jaladh-singhal:
> those could maybe go into a new test md file in the plugin itself, or added as a test md file if/when the plugin will live in its separate repo)

@bsipocz Agreed. I plan to add a readme when it becomes more mature and independent. That readme will be a good place for these test cases.

myst.yml
bsipocz: I am definitely thinking of multiple plugins, whether to ship them one per repo or multiple in a plugins one is a separate question :)

tox.ini
bsipocz: Haha, yes, we do use debug level logging to have more info available when there are issues. 

Ideally no content executing testing was done with these changes as you haven't touched the notebooks themselves; so I suppose we should go back and check what was going on with that failure.

bsipocz:
OK, so features that comes to mind, but I don't think we should wait for any of them and just get this PR merged as is already: 

- being able to wildcard the list of notebooks for the gallery. I like the current explicit approach, but I think it would make sense to be able to also say `tutorials/euclid/*md`
- alternatively to the wildcard, and this is very IRSA specific, we could also match on the field `'section'`.

- we could raise an error for the unrecognised notebooks when using ``--strict``. I'm not sure how complicated this would be.

- have the test cases moved into the plugin directory/future repository; as you suggested a README would suffice to start with as it could both serve as documentation and test

- probably a separate plugin/widget idea; or maybe an add-on feature here: do add a filtering on top of the gallery view. This would definitely be a feature that would also be super useful for upstream.

troyraen (approved):
This is great, thank you!

tox.ini
jaladh-singhal: I reverted it back. I think it's failing because of a bad DOI:
```
⛔️ tutorials/techniques-and-tools/SEDs_in_Firefly.md:93 Citation data from doi.org was not available or malformed for doi:10.3847/1538-3881/ace32f

   To resolve this error, visit https://doi.org/10.3847/1538-3881/ace32f and add citation info to local BibTeX file

To suppress this message, add rule: "doi-link-valid" with key: "https://doi.org/10.3847/1538-3881/ace32f#ajace32ff10" to "error_rules" in your project config
```
Or something else? I can't tell from:
```
Site has 1 error and 0 warning, stopping build.
py313-buildhtml: exit 1 (10.55 seconds) /home/circleci/project> bash -c 'jupyter-book build --execute --execute-parallel 1 --html --strict -d' pid=645
  py313-buildhtml: FAIL code 1 (94.58=setup[82.21]+cmd[0.43,0.00,0.04,0.04,0.00,0.00,0.00,0.01,0.36,0.01,0.00,0.09,0.00,0.83,10.55] seconds)
  evaluation failed :( (94.63 seconds)
```

bsipocz: Yes, the error is due to the DOI, but I don't think it's an actual bad DOI, but instead the server acts up occasionally.  Either way, it's unrelated to this PR.

bsipocz (approved):
Thank you so much again!

This all looks good now, thank you! I've restarted the preview, let's wait for it to double check, but if it keeps failing with unrelated issues I'll just go ahead and merge it without the preview.

tox.ini
jaladh-singhal: Thanks. Just curious: so rebuilding fixed it on its own?

bsipocz: I see this error often, both with the DOI links and other links; and they are heisenbug types. I guess we ultimately run into some rate limits or other DoS preventions on the server side. 

So yes, usually a restart resolves these types of issues; but I have also reported upstream a few variants that are more prevalent, like https://github.com/jupyter-book/mystmd/issues/2209

PR #307, opened 2026-04-30
Simplify & update firefly_client use in Euclid PHZ tutorial

jaladh-singhal:
I made these fixes recently when demoing the notebook on Fornax to Euclid HQ folks. Removes use of deprecated methods, attributes.

### DIFF
--- a/tutorials/euclid/4_Euclid_intro_PHZ_catalog.md
+++ b/tutorials/euclid/4_Euclid_intro_PHZ_catalog.md
@@ -73,6 +73,7 @@ We rely on ``astroquery`` features that have been recently added, so please make
 ```{code-cell} ipython3
 import os
 import re
+import io
 import urllib
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


tutorials/euclid/4_Euclid_intro_PHZ_catalog.md
```diff
@@ -394,51 +395,33 @@ ax.imshow(cutout_galaxy.data, cmap='gray', origin='lower',
...
-Save the data locally if you have not already done so, in order to upload to IRSA viewer.
-
 ```{code-cell} ipython3
-download_path = "data"
-if os.path.exists(download_path):
-    print("Output directory already created.")
-else:
-    print("Creating data directory.")
-    os.mkdir(download_path)
+fc = FireflyClient.make_client('https://irsa.ipac.caltech.edu/irsaviewer')
```
bsipocz, line 401: I would move this code cell down, right after current line 404. (I like having the narrative first and the see the code that does what the narrative describes)

tutorials/euclid/4_Euclid_intro_PHZ_catalog.md
```diff
@@ -394,51 +395,33 @@ ax.imshow(cutout_galaxy.data, cmap='gray', origin='lower',
...
 ```
 
-### Vizualize the image with Firefly
+### Vizualize the FITS image with Firefly
 
-First initialize the client, then set the path to the image, upload it to firefly, load it and align with WCS.
+After initializing the client, we can directly use the FITS image file URL to upload it to Firefy.
```
bsipocz: Unrelated to this PR; but I think we should cross link from firefly using notebooks to a firefly intro notebook. E.g. reading these updates from the POV of an e.g. fornax users of what you mean by "upload", if and how much of data transfer happens between what computers.

So, while all of that narrative would be nice, I don't think that we should repeat it in every notebook, but at minimum we should point to some docs or another notebook that answers those potential questions.

bsipocz (approved):
Look good, but I have some linking suggestion that you may consider for a follow-up PR that touches all the firefly notebooks.

tutorials/euclid/4_Euclid_intro_PHZ_catalog.md
jaladh-singhal: I agree - we have too long text in other irsa-tutorials about initializing or too little text (like in this one). I'll make a ticket to standardize firefly initializing across notebooks and cross-linking a dedicated firefly-initializing notebook we have here: https://caltech-ipac.github.io/firefly_client/usage/initializing-vanilla.html

jaladh-singhal: see https://github.com/Caltech-IPAC/irsa-tutorials/issues/310

PR #320, opened 2026-05-26
MAINT: Adding upstream optional dependency to install ist

### LINKED ISSUE #318: Euclid Cluster notebook seems to need an extra install
Running the latest version on my laptop it complains about not finding lxml so I had to do 

!pip install lxml

can be added to the first cell.

bsipocz:
Resolves #318


bsipocz:
The package to be installed is picked up properly, the CI status is unrelated and runs into the planned server side downtime. Merging the PR.

PR #325, opened 2026-05-27
Create tutorial to get ZTF light curves from HATS catalog

jaladh-singhal:
Fixes [IRSA-7768](https://jira.ipac.caltech.edu/browse/IRSA-7768)

PR #331, opened 2026-06-03
Upgrade lsdb to >=0.8

### LINKED ISSUE #197: Update notebooks for lsdb v0.8 `crossmatch()` changes
lsdb v0.8 will be released soon. The crossmatch() function will include a breaking change. The default suffix behavior will change from applying suffixes to all columns to only applying suffixes to overlapping columns. The following notebooks use that function:

- irsa-hats-with-lsdb.md

Several column names will need to be changed.

jaladh-singhal:
I chose 0.8.1 because that's what we're using in fornax-demo-notebooks and it was a hotfix for problems in 0.8.0 (https://github.com/astronomy-commons/lsdb/releases/tag/v0.8.1)

Closes #197

### DIFF
--- a/tutorial_requirements.txt
+++ b/tutorial_requirements.txt
@@ -12,7 +12,8 @@ pyarrow>=10.0.1
 hpgeom
 pandas[xml]>=1.5.2
 dask[distributed]
-lsdb>=0.6.6,<0.8
+# lsdb<0.8 returns wrong results for some catalogs, e.g. ZTF DR24 Lightcurves
+lsdb>=0.8.1
 psutil
 ray
 s3fs
--- a/tutorials/techniques-and-tools/irsa-hats-with-lsdb.md
+++ b/tutorials/techniques-and-tools/irsa-hats-with-lsdb.md
@@ -392,9 +392,10 @@ Since ZTF objects are defined per band, setting `n_neighbors=1` means this is on
 ```{code-cell} ipython3
 euclid_x_ztf = euclid_cone.crossmatch(
     ztf_cone,
-    suffixes=("_euclid", "_ztf"), # to distinguish columns from the two catalogs
     n_neighbors=1, # default is 1 too, can be tweaked
-    radius_arcsec=1  # default is 1 arcsec too, can be tweaked
+    radius_arcsec=1,  # default is 1 arcsec too, can be tweaked
+    suffixes=("_euclid", "_ztf"), # to distinguish columns from the two catalogs
+    suffix_method="all_columns",
 )
 euclid_x_ztf
 ```
--- a/tutorials/techniques-and-tools/irsa-hats-with-lsdb.md
+++ b/tutorials/techniques-and-tools/irsa-hats-with-lsdb.md
@@ -56,7 +56,7 @@ We will use lsdb to leverage HATS partitioning for performing fast spatial queri
 
 ```{code-cell} ipython3
 # Uncomment the next line to install dependencies if needed.
-# !pip install s3fs "lsdb>=0.6.6,<0.8" pyarrow pandas numpy astropy dask matplotlib
+# !pip install s3fs "lsdb>=0.8.1" pyarrow pandas numpy astropy dask matplotlib
 ```
 
 ```{code-cell} ipython3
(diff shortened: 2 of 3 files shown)


tutorials/techniques-and-tools/irsa-hats-with-lsdb.md
```diff
@@ -562,10 +563,17 @@ ztf_lcs
...
+Note: You may see a memory warning from lsdb which is expected due to the large size of Lightcurve data.
 
 ```{code-cell} ipython3
-ztf_lcs_df = ztf_lcs.compute() # ID search runs out of memory if we try to parallelize it with Dask client
+with Client(n_workers=get_nworkers(ztf_lcs),
+            threads_per_worker=1,
+            memory_limit=None # to prevent it from running out of memory
+            ) as client:
+    print(f"This may take more than a few minutes to complete. You can monitor progress in Dask dashboard at {client.dashboard_link}")
+    ztf_lcs_df = ztf_lcs.compute()
```
troyraen: This is causing circleci to run out of memory and fail. I'll try running it when I'm back at a computer.

troyraen (commented):
irsa-hats-with-lsdb.md
ztf_lc_df = ztf_lc.compute()

tutorials/techniques-and-tools/irsa-hats-with-lsdb.md
bsipocz: RAM usage seems to peak a few times in the latest run, but I have never seen this dask traceback before, so it's a bit different right now.

troyraen: I tried a few different things to try to reduce the memory usage of this call. Running it without a client used the least for me (5.5G max). It seems in the other notebook we found that the cumulative effect of multiple `.compute()` calls put us over the limit, so we needed to use a client for the index call so that it would release the memory, but here it's the added overhead of using a client that puts us over the limit.

FWIW, the only other ways I found to keep this call under the 8G limit of our CircleCI machine were to use `nworkers=1` (7.3G max) or reduce to very minimal `columns = ['objectid', 'lightcurve.mag', 'lightcurve.hmjd']` (7.7G max). I think both of those are worse options for the end user.

bsipocz: ok, let's skip this on the circleCI job then and see if it passes on GHA. -- I'm planning to drop circleCI altogether anyway in the near future.

troyraen: If we just remove the client here so that this notebook uses a max of 5.5G, will that be enough to leave in the circleci job?

bsipocz: I think that may just work, but we may see it easier if you just push a commit and try

troyraen: @jaladh-singhal can you remove this `client` so we can see if that's enough to get the notebook to pass circleCI?

jaladh-singhal: Sorry this got sidetracked after AAS since I was only focusing on the ZTF LC PR. Just pushed a commit - let's see.

jaladh-singhal: Good news, circleCI passed. Although we see memory warnings which is expected and narrative also mentions the same.

troyraen (approved):
Looks good to me now! The Buildhtml job got stuck, so I just restarted it. If/when that passes, I think it's ready to merge.

jaladh-singhal:
Build-html keeps getting stuck. Maybe we can just merge or wait until Brigitta is back?

troyraen:
Jaladh and I talked over zoom and decided to merge this. Most of the failures are happening because the job just silently hangs with no error or traceback, and it happens on a different notebook each time. This happens with these jobs sometimes and I think it indicates a network connection issue on the GHA backend. I don't _think_ we're directly causing it or can easily control it. I definitely don't think it's caused by this PR.

I have some concern about merging because this job isn't one we typically run on PRs; it was explicitly triggered when Brigitta added the "GHA buildhtml" label, probably because she wanted to force all notebooks to execute to make sure the change to dependencies didn't break one of them. But any breakage would only be to our own CI. The only user-facing change is to the HATS LSDB notebook, and that's been tested. So let's merge so Jaladh can use this in #325 and we'll check for breakages to other notebooks separately.

tutorial_requirements.txt
```diff
@@ -12,7 +12,8 @@ pyarrow>=10.0.1
 hpgeom
 pandas[xml]>=1.5.2
 dask[distributed]
-lsdb>=0.6.6,<0.8
+# lsdb<0.8 returns wrong results for some catalogs, e.g. ZTF DR24 Lightcurves
```
bsipocz, line 15: one minor comment here: if lsdb <0.8 has the problem then why we don't use that as a pin rather than 0.8.1? If there is any other reasons to go with the newer version as minimum, that should be consistently stated with the comment.

jaladh-singhal: This was deliberate as I described in the PR description:
> I chose 0.8.1 because that's what we're using in fornax-demo-notebooks and it was a hotfix for problems in 0.8.0 (https://github.com/astronomy-commons/lsdb/releases/tag/v0.8.1)

Re adding it in the comment, I didn't do it to keep the comment small. If you think it's important enough, we can add it in the next iteration.

bsipocz: yes, I think consistency pays out long term. We don't need to use 0.8, but our problem was fixed by it and not by 0.8.1. As the upstream team didn't yank the 0.8 release that is that one that actually matter for us. If it's buggy enough to be not usable, then we should move onto 0.8.1

(otherwise we could always just pin the latest in a bugfix release branch but that is kind of counter productive with how "minimum" should work.)

PR #326, opened 2026-05-29
reducing memory usage on OU TDE notebook

jkrick:
This PR reduces memory usage in the TDE light curve notebook to address CI failures likely caused by memory constraints.  It does not effect the science or goals of the notebook 

  Changes:
  - Replace full-image background estimation (np.nanmedian/np.nanstd over 4088×4088 pixels) with a local annulus-based background using photutils SkyCircularAnnulus and ApertureStats
  - Add del data in show_gallery() to free each full image array immediately after display
  - Add del img in make_cutout() to free the full image array after the cutout is extracted
  - Reduce default number of gallery images from 6 to 4
  - Reduce default number of cutout images from 6 to 4

PR #335, opened 2026-06-17
Hide cells both in jupyterlab and myst rendering

bsipocz:
Experimenting with a bit back and forth with Angus, I found a solution that apparently works for both usages we have. Draft as I still need to run a grep/sed to change all. And will also need to follow-up with a bit of documentation cleanup in the mystmd docs.

#myst-education-2026

closes https://github.com/IPAC-SW/irsa-datascience/issues/171

### DIFF
--- a/tutorials/euclid/merged-objects-hats-catalog/4-euclid-q1-hats-magnitudes.md
+++ b/tutorials/euclid/merged-objects-hats-catalog/4-euclid-q1-hats-magnitudes.md
@@ -232,6 +232,7 @@ Since the template-fit photometry is recommended for extended objects, we'll sep
 ---
 jupyter:
   source_hidden: true
+tags: [hide-cell]
 ---
 # Galaxy + any. Star + galaxy. QSO + galaxy.
 classes = {"Galaxy": (2, 3, 6, 7), "Star": (1, 3), "QSO": (4, 6)}
@@ -304,6 +305,7 @@ This figure is inspired by Romelli Fig. 6 (top panel).
 ---
 jupyter:
   source_hidden: true
+tags: [hide-cell]
 ---
 # Only consider objects within these mag and mag difference limits.
 mag_limits, mag_diff_limits = (16, 24), (-1, 1)
--- a/tutorials/euclid/euclid_clusters_tutorial.md
+++ b/tutorials/euclid/euclid_clusters_tutorial.md
@@ -188,6 +188,7 @@ To explore a different control field, change the seed value or remove the seed e
 ---
 jupyter:
   source_hidden: true
+tags: [hide-cell]
 ---
 def check_mer_tile_requirement(coord, search_radius=2.0):
     """Check whether a sky coordinate is covered by exactly 4 Euclid MER science images.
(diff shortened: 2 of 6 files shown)


bsipocz:
Test failures are not related to changes here and this is ready for review and merge. 

With the review I mean @jkrick or @troyraen - please have a look at the rendering to see if you are happy with the hidden cells, e.g. in here: https://circle.scientific-python.dev/output/job/701f33d8-d06c-4171-9a38-d0054519489e/artifacts/0/_build/html/openuniverse2024-tde-light-curve

Other option is that we can hide input or we can hide output, right now I went with hiding the whole cell.

jkrick:
Thanks for working on this.  I think this is an improvement over having the code cells in the notebook.  I don't like that the hidden cell in the rendering says "Notebook ...".  Do we have the ability to change that text phrase to be something of our choosing?  ie., "Function Definitions" or "Code Cell" or ??

tutorials/euclid/merged-objects-hats-catalog/4-euclid-q1-hats-magnitudes.md
```diff
@@ -232,6 +232,7 @@ Since the template-fit photometry is recommended for extended objects, we'll sep
 ---
 jupyter:
   source_hidden: true
+tags: [hide-cell]
```
troyraen:
```suggestion
tags: [hide-input]
```

tutorials/euclid/merged-objects-hats-catalog/4-euclid-q1-hats-magnitudes.md
```diff
@@ -304,6 +305,7 @@ This figure is inspired by Romelli Fig. 6 (top panel).
...
```
troyraen:
```suggestion
tags: [hide-input]
```

troyraen (commented):
Thanks @bsipocz! This will be helpful.

I see a couple cells that produce figures so I feel they should use `hide-input` instead of `hide-cell` so the figures are still displayed by default. I'll mark them. But actually I don't see a reason not to just use `hide-input` for everything.

> hidden cell in the rendering says "Notebook ...". Do we have the ability to change that text phrase to be something of our choosing?

I agree it would be great to customize the text if possible so the user sees something more informative than "Notebook Cell". I checked docs and found that according to [myst-nb docs](https://myst-nb.readthedocs.io/en/latest/render/hiding.html#hide-code-cells) (which might not be relevant), it might be possible with something like:

```
---
jupyter:
  source_hidden: true
tags: [hide-cell]
mystnb:
  code_prompt_show: "Show code"
  code_prompt_hide: "Hide code"
---
```

That doesn't work for me when building locally, but I'm building with `myst start` so it may still work in production with `npm`.

bsipocz:
> I see a couple cells that produce figures so I feel they should use hide-input instead of hide-cell so the figures are still displayed by default. I'll mark them. But actually I don't see a reason not to just use hide-input for everything.

sure, I was very much on the fence if it should be hide-cell or hide-input, so will just change everything to hide-input.


> I agree it would be great to customize the text if possible so the user sees something more informative than "Notebook Cell". I checked docs and found that according to [myst-nb docs](https://myst-nb.readthedocs.io/en/latest/render/hiding.html#hide-code-cells) (which might not be relevant), it might be possible with something like:

We don't use myst-nb, that is part of the old JB1 stack

bsipocz:
Afaik there is no way currently to customise the text, but I made a note about it and will follow-up upstream.

bsipocz:
I'm going ahead and merge this now. All the test failures are unrelated and are due to some yet unchartered problem with the TDE notebook. The same errors are present in the more relevant PR #326, so we should focus on fixing them there or open a separate issue for it.

Issue #315, opened 2026-05-20
CI: resolve fsps install issues for rendering CI runs and enable execution for the OU SED notebook


bsipocz:
See the log:

```
Building wheels for collected packages: fsps, pims
  Building wheel for fsps (pyproject.toml) ... error
  Building wheel for pims (pyproject.toml) ... done
  Created wheel for pims: filename=pims-0.7-py3-none-any.whl size=84644 sha256=0d5823bc0f102c81d56abc3e78a57619896628426aa58b2a2307b0f7fe8eba88
  Stored in directory: /home/circleci/.cache/pip/wheels/9b/9e/a8/a0f41cf12a4a7b98a501ad8590595e2071a946d2fd33b0eb60
Successfully built pims
Failed to build fsps
  error: subprocess-exited-with-error
  
  × Building wheel for fsps (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [25 lines of output]
      WARNING: Use cmake.version instead of cmake.minimum-version with scikit-build-core >= 0.8
      WARNING: Use ninja.version instead of ninja.minimum-version with scikit-build-core >= 0.8
      *** scikit-build-core 0.12.2 using CMake 3.22.1 (wheel)
      *** Configuring CMake...
      loading initial cache file /tmp/tmp_ptgjo82/build/CMakeInit.txt
      -- The C compiler identification is GNU 11.4.0
      -- The Fortran compiler identification is unknown
      -- Detecting C compiler ABI info
      -- Detecting C compiler ABI info - done
      -- Check for working C compiler: /usr/bin/gcc - skipped
      -- Detecting C compile features
      -- Detecting C compile features - done
      CMake Error at CMakeLists.txt:2 (project):
        No CMAKE_Fortran_COMPILER could be found.
      
        Tell CMake where to find the compiler by setting either the environment
        variable "FC" or the CMake cache entry CMAKE_Fortran_COMPILER to the full
        path to the compiler, or to the compiler name if it is in the PATH.
      
      
      -- Configuring incomplete, errors occurred!
      See also "/tmp/tmp_ptgjo82/build/CMakeFiles/CMakeOutput.log".
      See also "/tmp/tmp_ptgjo82/build/CMakeFiles/CMakeError.log".
      
      *** CMake configuration failed
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for fsps
[...]

PR #347, opened 2026-07-06
Remove fsps from requirements

### LINKED ISSUE #346: `scikit-build-core` causing latest-deps job to fail
Eg, see #345. `scikit-build-core` released v1.0.0 just 5 hours ago, so I'm guessing that's what's causing this. Traceback looks like this:

```
py313-test-devdeps: install_deps> bash -c 'if python -c '"'"'import platform; print(platform.platform())'"'"' | grep -i win; then sed -i -e '"'"'s|fsps|#fsps|g'"'"' tutorial_requirements.txt; elif [[ $CIRCLECI == true ]]; then sed -i -e '"'"'s|fsps|#fsps|g'"'"' tutorial_requirements.txt; fi && python -I -m pip install $@' 'astropy>0.0.dev0' pip git+https://github.com/astropy/astroquery.git git+https://github.com/astropy/pyvo.git -r test_requirements.txt -r site_requirements.txt -r tutorial_requirements.txt
  Running command git clone --filter=blob:none --quiet https://github.com/astropy/astroquery.git /tmp/pip-req-build-0gg193fa
  Running command git submodule update --init --recursive -q
  Running command git clone --filter=blob:none --quiet https://github.com/astropy/pyvo.git /tmp/pip-req-build-sjzjxhbd
  error: subprocess-exited-with-error
  
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 7
  ╰─> [1 lines of output]
      ERROR: Use cmake.version instead of cmake.minimum-version with scikit-build-core >= 0.8
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
ERROR: Failed to build 'fsps' when getting requirements to build wheel
```

troyraen:
Closes #346

### DIFF
--- a/tutorial_requirements.txt
+++ b/tutorial_requirements.txt
@@ -29,8 +29,6 @@ scikit-learn
 gdown
 astro-sedpy
 astro-prospector
-dynesty<2
-fsps
 seaborn
 regions
 # For supporting myst-based notebooks
--- a/tutorials/simulated-data/OpenUniverse2024/openuniverse2024_SED_fit.md
+++ b/tutorials/simulated-data/OpenUniverse2024/openuniverse2024_SED_fit.md
@@ -68,6 +68,7 @@ starttime = time.time()
 # Uncomment to install dependencies if needed.
 # %pip install numpy pandas h5py matplotlib seaborn pyarrow gdown
 # FSPS + Prospector + SEDPY+ dynasty are required for SED modeling
+# If you have trouble installing FSPS, you can skip SED modeling.
 # %pip install astro-prospector astro-sedpy "dynesty<2.0.0" fsps
 ```
 


troyraen:
I added the `GHA buildhtml` label to check whether notebooks still run but that job isn't executing. Tried triggering it manually after adding the label but still no luck. Does it require pushing a new commit after the label has been applied?

bsipocz (changes requested):
I think we should instead:

- limit the scikit-build version number as we'll run into this very same problem again when we enable fsps OR
- remove all the new dependencies brought in by that notebook -- so it's all consistent

At minimum we should maybe consider commenting the dependency out with a note that it's temporary and is due to scikit-build. Bottom line, it's OK to do a quick patch to fix up seemingly unrelated failures, but we should make our life easier to reverse it back when the workaround is expected to be needed for a longer period of time (e.g. I don't expect fsps will be new scikit-build compatible any time soon, if ever)

troyraen:
Sorry, I thought this notebook was never executed based on #315 and that fsps was just left in the requirements file by mistake. Looking closer at the CI config files I see that it is executed in some cases. I'll pin scikit-build-core.

bsipocz (dismissed):

bsipocz (approved):
I would think we could add a test job that opts in the SED fitting, but that should be separate from this PR.

troyraen:
For future reference:

`fsps` doesn't provide a wheel for python 3.13, so we end up building it from source. This ignores our requirements file and, since their toml doesn't pin scikit-build-core, it picks up latest which fails as of yesterday's scikit-build-core v1.0.0 release.

We could work around that, but we discovered that the OpenUniverse SED notebook doesn't use `fsps` by default (requires user to change `RUN_FITS` to `True`), so there's no need for our testing and deployment infrastructure to install it.

bsipocz:
Upstream report: https://github.com/dfm/python-fsps/issues/388

(but I think we should stick with python 3.12 for the fsps enabled case)

troyraen:
Ah, thanks for reporting upstream. Realized I didn't do it while trying to fall asleep last night.

PR #354, opened 2026-07-10
CI: setting up usage of JB execution cache for GHA deploy job

bsipocz:
I don't expect to see much in this PR as the job in question is not triggered for PRs, but nevertheless make the changes in a PR for visibility and transparency.

~The TMP commit needs to be removed before merging.~ done


bsipocz:
This should be good to go, no need to wait for CI as the changes has no effect on the test matrix.

PR #364, opened 2026-08-03
SPHEREx PSF notebook update using ePSf

afaisst:
Closes https://github.com/IPAC-SW/irsa-datascience/issues/225

update to use of new ePSF.
Currently, data is not public, so the notebook is run on local files.

Leaving this as a draft PR until the data is public so it's not accidentally merged before then.

### DIFF
--- a/tutorials/spherex/spherex_psf.md
+++ b/tutorials/spherex/spherex_psf.md
@@ -3,14 +3,14 @@ authors:
 - name: Vandana Desai
 - name: Jessica Krick
 - name: Andreas Faisst
-- name: "Brigitta Sipőcz"
+- name: "Brigitta Sip\u0151cz"
 - name: Troy Raen
 jupytext:
   text_representation:
     extension: .md
     format_name: myst
     format_version: 0.13
-    jupytext_version: 1.18.1
+    jupytext_version: 1.17.2
 kernelspec:
   display_name: Python 3 (ipykernel)
   language: python
@@ -58,7 +58,7 @@ The following packages must be installed to run this notebook.
 
 ```{code-cell} ipython3
 # Uncomment the next line to install dependencies if needed.
-# !pip install astropy matplotlib numpy pyvo
+# !pip install astropy numpy pyvo
 ```
 
 ```{code-cell} ipython3
@@ -84,7 +84,12 @@ from astropy.utils.data import conf
 conf.remote_timeout = 120
 ```
 
-## 4. Get SPHEREx Cutout
+## 4. Get SPHEREx Cutout and Obtain Position on Parent Frame
+
+In this section, we show how to obtain a SPHEREx cutout and obtain the pixel position of the cutout on the parent LVF image.
+
+
+### 4.1 Obtain SPHEREx Cutout
 
 We first obtain a SPHEREx cutout for a given coordinate of interest from IRSA archive.
 For this we define a coordinate and a size of the cutout.
@@ -141,7 +146,7 @@ spectral_image_url = results['uri'][0]
 print(spectral_image_url)
 ```
 
-## 5. Read in a SPHEREx Cutout
+### 4.2 Read SPHEREx Cutout
 
 Next, we use standard astropy tools to open the fits image and to read the different headers and data.
 Transient read errors occur sometimes, so we'll catch those and retry a few times.
@@ -159,9 +164,7 @@ for attempt in range(max_retries):
         with fits.open(spectral_image_url) as hdul:
             image_hdul = copy.deepcopy(hdul)
             cutout_header = hdul['IMAGE'].header
-            psf_header = hdul['PSF'].header
             cutout = hdul['IMAGE'].data
-            psfcube = hdul['PSF'].data
         break
     except (TimeoutError, urllib.error.HTTPError, http.client.IncompleteRead):
         if attempt == max_retries - 1:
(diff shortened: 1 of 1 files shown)


tutorials/spherex/spherex_psf.md
```diff
@@ -58,11 +58,12 @@ The following packages must be installed to run this notebook.
 
 ```{code-cell} ipython3
 # Uncomment the next line to install dependencies if needed.
-# !pip install astropy matplotlib numpy pyvo
+# !pip install astropy numpy pyvo
```
troyraen: Put matplotlib back in here since it's used explicitly in the notebook.

tutorials/spherex/spherex_psf.md
```diff
@@ -115,7 +121,7 @@ service = pyvo.dal.TAPService("https://irsa.ipac.caltech.edu/TAP")
 query = f"""
 SELECT
     'https://irsa.ipac.caltech.edu/' || a.uri || '?center={ra.value},{dec.value}d&size={size.value}' AS uri,
-    p.time_bounds_lower
+    p.time_bounds_lower, energy_bandpassname, time_bounds_lower, time_bounds_upper, provenance_version
```
troyraen: Will the additional column names need `a.` or `p.` prepended?

tutorials/spherex/spherex_psf.md
```diff
@@ -159,481 +166,441 @@ for attempt in range(max_retries):
...
+Note that the ePSFs are stored in the form of a binary table, which makes it straight forward to obtain additional information on the ePSF (such as its location on the larger SPHEREx LVF image).
 
-    VERSION_FIXED = Version("6.5.6")
-    PSF_FIX_TAG = "psffix1"
+```{code-cell} ipython3
+try:
+    epsf_bintable = Table(image_hdul['EPSF'].data)
+    epsf_header = image_hdul['EPSF'].header
+    print("ePSF is loaded!")
+except Exception as e:
```
troyraen: We need to catch the specific error we're expecting here rather than using the very broad `Exception`. I can help figure out what that is if needed.

tutorials/spherex/spherex_psf.md
```diff
@@ -159,481 +166,441 @@ for attempt in range(max_retries):
...
+        psf_stamp : `numpy.ndarray`
+            Rendered PSF image with shape (``stamp_size``, ``stamp_size``).
+        x_ll, y_ll : `int`
+            Integer lower-left pixel index of the stamp in the image coordinate system.
+            That is, psf_stamp[0,0] corresponds to image pixel (y_ll, x_ll).
+        """
 
-psf_cube_plane = tab[0]["zone_id"]
-distance_min = tab[0]["distance"]
+        # Not sure why this was necessary
```
troyraen: We should try to understand this and put a more informative comment here. I can help with that if needed.

tutorials/spherex/spherex_psf.md
```diff
@@ -159,481 +166,441 @@ for attempt in range(max_retries):
...
-**Updated:** 25 March 2026
+**Updated:** 29 July 2026
 
 **Contact:** Contact [IRSA Helpdesk](https://irsa.ipac.caltech.edu/docs/help_desk.html) with questions or problems.
 
 **Runtime:** Approximately 30 seconds.
 
+```{code-cell} ipython3
+
+```
```
troyraen:
```suggestion
```

troyraen (commented):
I skimmed through it and left a few comments below.

tutorials/spherex/spherex_psf.md
```diff
@@ -159,481 +166,441 @@ for attempt in range(max_retries):
...
-        Rules:
-        - If the VERSION header is missing in the primary HDU, the fix is not applied.
-        - If VERSION >= VERSION_FIXED, the fix is included in the software release.
-        - Otherwise the local version tag (+...) must contain PSF_FIX_TAG.
-        """
-        header = hdul[0].header
+### 5.2 Obtaining the old PSF (versions $<$ 7.0)
+
+We advice the users **not** to use the older PSFs, which are stored in the `PSF` layer in the FITS file of the LVF images.
+Since in these older versions the ePSFs are not provided in the FITS images, we have to obtain them from the respective calibration products.
```
jkrick, line 291: Will the archive versions of all files be updated to include the ePSF?  is having a PSF only a temporary thing?  For the case where someone has downloaded their data already, but a newer version exists in the archive, could we say that users can re-download their data to get the newest version instead of this options?

tutorials/spherex/spherex_psf.md
```diff
@@ -159,481 +166,441 @@ for attempt in range(max_retries):
...
-    # Look for keys like XCTR* or YCTR*
-    xm = re.match(r'(XCTR*)', key)
-    if xm:
-        xplane = int(key.split("_")[1])
-        xctr[xplane] = val
-    ym = re.match(r'(YCTR*)', key)
-    if ym:
-        yplane = int(key.split("_")[1])
-        yctr[yplane] = val
+plt.show()
```
jkrick, line 479: This plot could use a figure caption that tells us what is interesting about this figure.  I think it is more useful to label the third y-axis, and the color bars as 'ePSF area" instead of "Neff", or even "PSF effective area", and what is that unit? pixels?

tutorials/spherex/spherex_psf.md
```diff
@@ -159,481 +166,441 @@ for attempt in range(max_retries):
...
 ```
 
-Once we have created this dictionary with zone pixel coordinates, we can simply search for the closest zone center to the coordinates of interest.
-For this we first add the distance between zone center coordinates and coordinates of interest to the table. 
-Note that the zone pixel center coordinates are 0-based, while their _names_ in the header (for example `XCTR_n`) are 1-based.
+Using the oversampling factor, we can then construct the ePSF Photutils model.
 
 ```{code-cell} ipython3
-tab["distance"] = np.sqrt((tab["x"] - xpix_orig)**2 + (tab["y"] - ypix_orig)**2)
+from photutils.psf import EPSFModel
```
jkrick: this should go near the top with the other imports.  Also, I ran into a problem with photutils not being installed.  It works to use `!pip install "photutils<3"`  but then I get a deprecation warning that I should be ising `ImagePSF` instead of photutils.

tutorials/spherex/spherex_psf.md
```diff
@@ -159,481 +166,441 @@ for attempt in range(max_retries):
...
+ax1 = fig.add_subplot(1,2,1)
+ax2 = fig.add_subplot(1,2,2)
+
+ax1.imshow(psf_stamp1, origin="lower")
+ax1.set_title("Source at (50,50)")
+
+ax2.imshow(psf_stamp2, origin="lower")
+ax2.set_title("Source at (100.5,100.5)")
+
+plt.show()
```
jkrick, line 621: please add a figure caption here as well.

jkrick (approved):
small changes requested, otherwise looks good.

tutorials/spherex/spherex_psf.md
afaisst: good call. it's the `p.` - added it.
It works without it... so either it is very smart or I got lucky

tutorials/spherex/spherex_psf.md
afaisst: changed it to catch a KeyError

tutorials/spherex/spherex_psf.md
afaisst: I think this is just to convert them to floats.
According to chatGPT, it suggests to just remove the comment and replace it with something like
```
# Convert to Python floats (accepts ints and NumPy scalar types).
```

Alternatively, we could do the test:
```
try:
    x0 = float(x0)
    y0 = float(y0)
except (TypeError, ValueError) as err:
    raise TypeError("x0 and y0 must be real numbers.") from err
```

What do you think? For now I just did option 1 (remove comment and replace it)

tutorials/spherex/spherex_psf.md
afaisst: not sure what's the difference?

tutorials/spherex/spherex_psf.md
afaisst: All the images on the archive will be replaced by the reprocessing (over the next couple of months). However, as you say, users might have older images that do not contain the ePSF. In principle, it would be advisable if those users download the images again (because of other calibrations). But for the PSF specifically, it wouldn't matter because the PSF is independent of the image (the actual PSF is the same for old and new image, just our measurement/modeling changed).

afaisst: I tried to put this in text. I added a note

tutorials/spherex/spherex_psf.md
afaisst: added to the pip install.
added import to the top.

afaisst: also changed to ImagePSF()

afaisst:
All comments above taken into account. @troyraen please have another look at the Exception comment you posted above to see if that makes sense.
Waiting for the data to be come public.

To do: 
- need to change the link from internal files to files on archive (or cloud).

tutorials/spherex/spherex_psf.md
troyraen: great!

tutorials/spherex/spherex_psf.md
troyraen: Good choice to go with option 1. Looking at how this gets called in the notebooks, it seems like the conversion shouldn't be necessary at all. But doing it doesn't hurt anything (just adds more lines of code 🙃 ).

tutorials/spherex/spherex_psf.md
troyraen: Sorry, that's probably confusing if you're not used to how github does things. The suggestion was to remove the extra, empty cell. Looks like you already did!

afaisst:
@troyraen notebook is updated to now get the ePSF calibration from the qr3 AWS s3 bucket. It seems to work, however, I couldn't run the full notebook as the SPHEREx TAP query is currently down.

tutorials/spherex/spherex_psf.md
```diff
@@ -159,481 +168,478 @@ for attempt in range(max_retries):
         with fits.open(spectral_image_url) as hdul:
             image_hdul = copy.deepcopy(hdul)
```
troyraen, line 169: @DeekshaMohanty we should probably use this deepcopy method in https://github.com/IPAC-SW/irsa-parquet/pull/116/

tutorials/spherex/spherex_psf.md
```diff
@@ -159,481 +168,478 @@ for attempt in range(max_retries):
...
+We advice the users **not** to use the older PSFs, which are stored in the `PSF` layer in the FITS file of the older (versions $<$ 7.0) LVF images.
+Since in these older versions the ePSFs are not provided in the FITS images, we have to obtain them from the respective calibration products. 
 
-    Return
-    ----------
-    new_hdul : fits.HDUList
-        New SPHEREx Spectral Image HDUL with updated PSF zone data in header and updated version number
-    """
+```{note}
+Eventually, after the reprocessing of the data products, the older version LVF images will be replaced by newer versions. Generally, we advise to download the newest images (which also contain improved flagging and calibrations). However, note that the effective PSF of the images is unchanged, but the ePSF provides a better modeling of it than the previous PSFs. It is therefore safe to just exchange the PSFs and use the new ePSF for the older version images.
```
troyraen, line 287: Good blurb for https://github.com/Caltech-IPAC/spherex-archive-documentation/pull/28

tutorials/spherex/spherex_psf.md
DeekshaMohanty: Okay! I can add this in.

troyraen (commented):
TAP issues have resolved and this is running fine for me on Fornax. I'll run a couple CI tests then merge.

Thanks @afaisst!

troyraen:
CircleCI is failing because of two notebooks not touched by this PR. The failures might be real problems but they're unrelated to this PR. So I'll chase those separately. Weird that CircleCI is executing untouched notebooks; need to look into that, too.

I pushed a commit with a little fix for string termination. The CircleCI job is trying to execute every notebook again, so I'm cancelling that. Will let the other tests go.

PR #365, opened 2026-08-11
Fixing the hidden code and figure cells in the Euclid q1 hats magntitude webpage

### LINKED ISSUE #361: Figures are being hidden by default
The figures in https://caltech-ipac.github.io/irsa-tutorials/euclid-q1-hats-magnitudes/ are being hidden in drop-down cells that are closed by default. Those should be using `hide-input` rather than `hide-cell` so the code is still hidden by default but the output figure is always displayed. left over from #335

DeekshaMohanty:
## Summary

This PR fixes the hidden code and figure cells in the Euclid Q1 HATS magnitudes tutorial page.

The figure outputs were being placed inside cells using `hide-cell`, which caused both the code and output figure to be hidden by default. These have been changed to `hide-input` so that only the figure output is displayed.

This issue was left over from #335.

## Related issue

Closes #361.

## Testing procedure

- Created an environment using `tutorial_requirements.txt` and `site_requirements.txt`.
- Built the documentation locally with Jupyter Book using the execution and strict-build options.
- Served the generated HTML locally and verified that the figure is displayed and the code remains hidden.

### DIFF
--- a/tutorials/euclid/merged-objects-hats-catalog/4-euclid-q1-hats-magnitudes.md
+++ b/tutorials/euclid/merged-objects-hats-catalog/4-euclid-q1-hats-magnitudes.md
@@ -232,7 +232,7 @@ Since the template-fit photometry is recommended for extended objects, we'll sep
 ---
 jupyter:
   source_hidden: true
-tags: [hide-cell]
+tags: [hide-input]
 ---
 # Galaxy + any. Star + galaxy. QSO + galaxy.
 classes = {"Galaxy": (2, 3, 6, 7), "Star": (1, 3), "QSO": (4, 6)}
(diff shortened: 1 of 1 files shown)


troyraen (approved):
Looks good. Thank you!

troyraen:
@DeekshaMohanty go ahead and merge this. It will trigger a deployment job on GHA (I think it's called "Build and Publish HTML"). Go to the Actions tab for this repo and keep an eye on that job. If it succeeds it will probably take between 10-30 minutes. If it takes much longer than that it likely means it's stuck. If it does that, or if it fails, let me know and we'll take a look at it together to see what went wrong.

PR #371, opened 2026-08-17
Align sidebar titles with notebook metadata

### LINKED ISSUE #300: Reconcile sidebar titles with mission index page titles
Now that we are generating notebook gallery from `notebook_metadata.yml` (#298), some of the short notebook titles are out of sync with the titles we have in the sidebar from `toc.yml`

We have 5 mission index pages, following notebooks' short tiles mismatch:
1. SPHEREx: None
2. Euclid:
    - almost half of the notebooks
    - HATS notebook will require special handling since they are nested
3. WISE:
   - AllWISE Source Catalog,
   - NEOWISE Solar System Objects
4. Simulated Data: None
5. Techniques & Tools:
   - HATS with LSDB

DeekshaMohanty:
## Summary

This PR updates the sidebar titles in `toc.yml` to match the titles in `notebook_metadata.yml`. I also flattened the nested Euclid Merged Objects Catalog so that the sidebar titles match the corresponding gallery cards. This closes #300.

### DIFF
--- a/toc.yml
+++ b/toc.yml
@@ -31,13 +31,15 @@ project:
           file: tutorials/euclid/5_Euclid_intro_SPE_catalog.md
         - title: Galaxy Clusters
           file: tutorials/euclid/euclid_clusters_tutorial.md
-        - title: 'Merged Objects Catalog: Introduction'
-          file: tutorials/euclid/merged-objects-hats-catalog/1-euclid-q1-hats-intro.md
-        - title: 'Merged Objects Catalog: Magnitudes'
-          file: tutorials/euclid/merged-objects-hats-catalog/4-euclid-q1-hats-magnitudes.md
+        - title: Merged Objects HATS Catalog
+          children:
+            - title: Introduction
+              file: tutorials/euclid/merged-objects-hats-catalog/1-euclid-q1-hats-intro.md
+            - title: Magnitudes
+              file: tutorials/euclid/merged-objects-hats-catalog/4-euclid-q1-hats-magnitudes.md
         - title: Euclid Cloud Access
           file: tutorials/euclid/euclid-cloud-access.md
-        - title: Early Release Observations (ERO) - Star Clusters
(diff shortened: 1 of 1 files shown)


toc.yml
```diff
@@ -31,24 +31,22 @@ project:
...
-            - title: Introduction
-              file: tutorials/euclid/merged-objects-hats-catalog/1-euclid-q1-hats-intro.md
-            - title: Magnitudes
-              file: tutorials/euclid/merged-objects-hats-catalog/4-euclid-q1-hats-magnitudes.md
-        - title: Cloud Access
+        - title: 'Merged Objects Catalog: Introduction'
+          file: tutorials/euclid/merged-objects-hats-catalog/1-euclid-q1-hats-intro.md
+        - title: 'Merged Objects Catalog: Magnitudes'
+          file: tutorials/euclid/merged-objects-hats-catalog/4-euclid-q1-hats-magnitudes.md
+        - title: Euclid Cloud Access
```
troyraen, line 40: We actually want to leave this nested like it was. Sorry the instructions weren't clear!

toc.yml
```diff
@@ -31,24 +31,22 @@ project:
...
           file: tutorials/euclid/euclid-cloud-access.md
-        - title: ERO Star Clusters
+        - title: Early Release Observations (ERO) - Star Clusters
```
troyraen: Let's leave this one like it was as well ("ERO Star Clusters"). Otherwise it's too long and has to wrap onto a second line. Sometimes we decide we need that, but in this case "ERO" is probably fine and folks who don't know what that is will find out when they click on it.

troyraen (approved):
Thanks! Flagging two things to revert because the instructions weren't clear (see below), then go ahead and merge.

toc.yml
DeekshaMohanty: Ah alright, good to know! I was wondering how to handle this and eventually decided to just flatten it - will change it back!

